PR #50, opened 2018-08-08
Beta sanitation

StevenLeighton21:
TODO - fix auth spec (mock is not working, auth is working fine)

- Prevent sending messages to actual phone numbers unless in a _properly configured_ production environment.
- Prevent user login unless ID token contains a matching group for configured IC_STAFF_GROUP

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -5,7 +5,8 @@ class SessionsController < ApplicationController
   def new; end
 
   def create
-    groups = auth_hash.extra.nil? ? nil : auth_hash.extra.raw_info.id_token
+    puts auth_hash
+    flash[:notice] = 'You do not have the required access permission' and return redirect_to login_path if auth_hash.extra.nil? || !check_active_groups
 
     user = find_or_create_user.execute(
       provider_uid: auth_hash.uid,
@@ -14,7 +15,7 @@ class SessionsController < ApplicationController
       email: auth_hash.info.email,
       first_name: auth_hash.info.first_name,
       last_name: auth_hash.info.last_name,
-      ad_groups: groups
+      ad_groups: auth_hash.extra.raw_info.id_token
     )
 
     session[:current_user] = {
@@ -24,7 +25,7 @@ class SessionsController < ApplicationController
       'groups_token' => user.fetch(:ad_groups)
      }
 
-    redirect_to root_path
+    redirect_to root_path and return
   end
 
   def destroy
@@ -34,6 +35,10 @@ class SessionsController < ApplicationController
   end
 
   private
+
+  def check_active_groups
+    auth_hash.extra.raw_info.id_token.split('.').include?(ENV['IC_STAFF_GROUP'])
+  end
 
   def auth_hash
     request.env['omniauth.auth']
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -16,14 +16,14 @@ class SessionsController < ApplicationController
       email: auth_hash.info.email,
       first_name: auth_hash.info.first_name,
       last_name: auth_hash.info.last_name,
-      ad_groups: auth_hash.extra.raw_info.id_token
+      provider_permissions: auth_hash.extra.raw_info.id_token
     )
 
     session[:current_user] = {
       'id' => user.fetch(:id),
       'name' => user.fetch(:name),
       'email' => user.fetch(:email),
-      'groups_token' => user.fetch(:ad_groups)
+      'groups_token' => user.fetch(:provider_permissions)
      }
 
     redirect_to root_path
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -5,8 +5,9 @@ class SessionsController < ApplicationController
   def new; end
 
   def create
-    puts auth_hash
+    # rubocop:disable Style/AndOr
     flash[:notice] = 'You do not have the required access permission' and return redirect_to login_path if auth_hash.extra.nil? || !check_active_groups
+    # rubocop:enable Style/AndOr
 
     user = find_or_create_user.execute(
       provider_uid: auth_hash.uid,
@@ -25,7 +26,7 @@ class SessionsController < ApplicationController
       'groups_token' => user.fetch(:ad_groups)
      }
 
-    redirect_to root_path and return
+    redirect_to root_path
   end
 
   def destroy
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -5,19 +5,23 @@ class SessionsController < ApplicationController
   def new; end
 
   def create
+    groups = auth_hash.extra.nil? ? nil : auth_hash.extra.raw_info.id_token
+
     user = find_or_create_user.execute(
       provider_uid: auth_hash.uid,
       provider: auth_hash.provider,
       name: auth_hash.info.name,
       email: auth_hash.info.email,
       first_name: auth_hash.info.first_name,
-      last_name: auth_hash.info.last_name
+      last_name: auth_hash.info.last_name,
+      ad_groups: groups
     )
 
     session[:current_user] = {
       'id' => user.fetch(:id),
       'name' => user.fetch(:name),
-      'email' => user.fetch(:email)
+      'email' => user.fetch(:email),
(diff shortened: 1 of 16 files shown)


spec/controllers/sessions_controller_spec.rb
```diff
@@ -25,9 +33,12 @@
       OmniAuth.config.mock_auth[:azure_activedirectory] = OmniAuth::AuthHash.new(
         provider: 'azure_activedirectory',
         uid: provider_uid,
-        info: info_hash
+        info: info_hash,
+        extra: extra_hash
       )
 
+      ENV['IC_STAFF_GROUP'] = '123456ABC'
```
zuren: Ideally we would never have to force global changes to a variable, like `ENV`, in a test. If this is strictly necessary, could we make sure we clean up afterwards instead of relying on another spec to do so?

For example, if randomised the order tests run in (as we should) this would start failing, or causing unexpected side effects, as the ENV var isn't cleared until the next spec AFAICS.

Preferably we'd either use an `after` block in this context:

```rb
after { ENV.delete('IC_STAFF_GROUP') }
```

Or alternative an `around` block:

```
around do |example|
  ENV['IC_STAFF_GROUP'] = '123456ABC'
  example.run
  ENV.delete('IC_STAFF_GROUP')
end
```

app/controllers/sessions_controller.rb
```diff
@@ -5,19 +5,25 @@ class SessionsController < ApplicationController
   def new; end
 
   def create
+    # rubocop:disable Style/AndOr
+    flash[:notice] = 'You do not have the required access permission' and return redirect_to login_path if auth_hash.extra.nil? || !check_active_groups
+    # rubocop:enable Style/AndOr
```
zuren: Could we just enable this globally in rubocop config? Feels dirty to disable it for one offs, but fine with disabling it if we think it's unreasonable linting.

app/controllers/sessions_controller.rb
```diff
@@ -5,19 +5,25 @@ class SessionsController < ApplicationController
...
+
     user = find_or_create_user.execute(
       provider_uid: auth_hash.uid,
       provider: auth_hash.provider,
       name: auth_hash.info.name,
       email: auth_hash.info.email,
       first_name: auth_hash.info.first_name,
-      last_name: auth_hash.info.last_name
+      last_name: auth_hash.info.last_name,
+      ad_groups: auth_hash.extra.raw_info.id_token
```
zuren: Using the terminology `ad_group` is now linking our implementation to Active Directory, which isn't ideal given we've been using `provider` so far. Is there a way we can make this generic and pass through parsed group objects as `provider_groups` or similar?

app/controllers/sessions_controller.rb
```diff
@@ -31,6 +37,10 @@ def destroy
 
   private
 
+  def check_active_groups
```
zuren: This method name isn't particularly descriptive, or obvious that it returns a boolean value.

Suggest we use something like `user_has_ic_staff_permissions?` instead?

config/database.yml
```diff
@@ -14,8 +14,8 @@ sqlite: &sqlite
   timeout: 5000
 
 development:
-  <<: *mysql
-  database: income_collection_development
+  <<: *sqlite
+  database: db/development.sqlite3
```
zuren: Probably preferable not to commit this solely due to local issues.

db/migrate/20180808095823_add_ad_group_ids_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddAdGroupIdsToUsers < ActiveRecord::Migration[5.2]
+  def change
+    add_column :users, :ad_groups, :string
```
zuren: Again, tying it to AD :(

lib/hackney/income/gov_notify_gateway.rb
```diff
@@ -38,6 +38,18 @@ def get_email_templates
           { id: template.id, name: template.name, body: template.body, subject: template.subject }
         end
       end
+
+      private
+
+      def pre_release_phone_number(phone_number)
+        phone_number if Rails.env.production? && ENV['SEND_LIVE_COMMUNICATIONS']
+        ENV['TEST_PHONE_NUMBER']
+      end
```
zuren: There's no test around this or the following method, and they won't work correctly AFAICS. `TEST_PHONE_NUMBER` and `TEST_EMAIL_ADDRESS` will be used in all cases as there's no explicit return on the first line.

app/controllers/sessions_controller.rb
StevenLeighton21: The rule says to replace `and` with `&&` which will break in this context

StevenLeighton21: But is a fine rule elsewhere

app/controllers/sessions_controller.rb
StevenLeighton21: I chose this way because we'd like add more providers rather than replace, right? So if we had a different provider, it would have its own way of managing access permissions which could mean another groups field specific to that provider.

app/controllers/sessions_controller.rb
zuren: Will it break in this context?

`if (auth_hash.extra.nil? and !check_active_groups)` with the parens would probably work?

lib/hackney/income/gov_notify_gateway.rb
StevenLeighton21: woops, good catch

app/controllers/sessions_controller.rb
zuren: I agree that we would. It would be nice if we had a common way of representing a group between all implementations. i.e.

```
provider_groups: [
  { provider: xxx, id: xxx, name: xxx, metadata: { any remaining provider specific stuff } } 
]
```

That way we wouldn't have to extend this use case over and over to support new providers, we'd just add more gateways which knew how to support the provider.

app/controllers/sessions_controller.rb
StevenLeighton21: It wants to replace the `and` in `and return`, which will break AFAIK (did when I tried it!)

spec/lib/hackney/income/gov_notify_gateway_spec.rb
```diff
@@ -14,10 +14,53 @@
...
+      subject.send_text_message(
+        phone_number: '12345',
+        template_id: 'sweet-test-template-id',
+        variables: {
+          'first name' => 'Steven Leighton',
+          'balance' => '-£100.00'
+        },
+        reference: 'amazing-test-reference'
+      )
+    end
```
zuren: Missing triangulation on this example - nothing to guarantee that what Notifications::Client receives isn't hardcoded.

spec/lib/hackney/income/gov_notify_gateway_spec.rb
```diff
@@ -14,10 +14,53 @@
...
+
+    after do
+      ENV['SEND_LIVE_COMMUNICATIONS'] = 'false'
```
zuren: Maybe delete the env var instead of resetting it to `'false'`?

app/controllers/sessions_controller.rb
zuren: Would probably go with:

```
if auth_hash.extra.nil? || !check_active_groups
  flash[:notice] = 'whatever'
  redirect_to login_path and return
end
```

Bit more readable and lets you do away with the nasty rubocop comments.

db/schema.rb
```diff
@@ -58,6 +60,8 @@
     t.string "email"
     t.string "first_name"
     t.string "last_name"
+    t.string "ad_groups"
+    t.string "provider_permissions"
```
zuren: Your schema now contains `ad_groups` and `provider_permissions`. Probably needed to rollback the penultimate migration rather than deleting it outright?

lib/hackney/income/gov_notify_gateway.rb
```diff
@@ -38,6 +38,22 @@ def get_email_templates
...
+
+      def pre_release_email(email)
+        return email if send_for_real?
+        ENV['TEST_EMAIL_ADDRESS']
+      end
+
+      def send_for_real?
+        ENV['SEND_LIVE_COMMUNICATIONS'] == 'true'
+      end
```
zuren: Great method name.

PR #53, opened 2018-08-15
Get case list in real time

StevenLeighton21:
These steps are temporary in nature in order to test live data. Future:
 - Map users to case lists on IC API
 - Requesting case list passes user ID as a param, is returned that user worktray

### DIFF
--- a/spec/lib/hackney/income/list_user_assigned_cases_spec.rb
+++ b/spec/lib/hackney/income/list_user_assigned_cases_spec.rb
@@ -87,7 +87,9 @@ describe Hackney::Income::ListUserAssignedCases do
       latest_action_date: Faker::Date.forward(100),
       primary_contact_name: [Faker::Name.prefix, Faker::Name.first_name, Faker::Name.last_name].join(' '),
       primary_contact_short_address: Faker::Address.street_address,
-      primary_contact_postcode: Faker::Address.postcode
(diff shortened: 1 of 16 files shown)


spec/lib/hackney/income/list_user_assigned_cases_spec.rb
```diff
@@ -2,19 +2,33 @@
...
+    it 'should return the cases as TenancyListItem instances with scores and bands' do
+      subject.each do |item|
+        expect(item).to be_instance_of(Hackney::Income::Domain::TenancyListItem)
+        expect(item.score).to_not eq(nil)
+        expect(item.band).to_not eq(nil)
+      end
+    end
+  end
+
+  xcontext 'when retrieving cases for a user who has none assigned' do
```
booyaa: Just to check this are still meant to be skipped or did they slip the commit?

StevenLeighton21: Yeah, they are for older sync functionality to be removed, I have added only a rudimentary test around the demo endpoint

StevenLeighton21: The reason is that the API I'm hitting currently has no concept of assigning data to users, yet.

PR #54, opened 2018-08-15
Remove MySQL config from database.yml

lukemorton:
Production and staging DB configuration should be read from DATABASE_URL

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -1,21 +1,11 @@
-mysql: &mysql
-  adapter: mysql2
-  encoding: utf8
-  username: <%= ENV.fetch("DATABASE_USERNAME") %>
-  password: <%= ENV.fetch("DATABASE_PASSWORD") %>
-  pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
-  host: <%= ENV.fetch("DATABASE_HOST") %>
-  port: <%= ENV.fetch("DATABASE_PORT") %>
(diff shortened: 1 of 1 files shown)


config/database.yml
```diff
@@ -1,21 +1,11 @@
...
-  port: <%= ENV.fetch("DATABASE_PORT") %>
-  # socket: <%# ENV.fetch("DATABASE_SOCKET") %>
-
 sqlite: &sqlite
   adapter: sqlite3
   pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
   timeout: 5000
 
 development:
-  <<: *mysql
```
StevenLeighton21: Should we use this permanently? It's a quick fix to get around docker issues when jumping between the projects.

lukemorton: If you ask me we should aim to get back to MySQL ASAP via Docker. Right now, we're all using SQLite in development and so lets update the app to reflect that.

config/database.yml
```diff
@@ -24,10 +14,4 @@ test:
...
   database: db/test.sqlite3
 
-staging:
-  <<: *mysql
-  database: income_collection_staging
-
-production:
-  <<: *mysql
-  database: income_collection_production
+# production should be injected via DATABASE_URL
```
StevenLeighton21: There's still a local database that stores the results of our priority calculations, I believe we shouldn't remove this? Or will it be configured via env _alongside_ the connection to UH?

lukemorton: This is injected via ENV.

PR #57, opened 2018-08-28
Remove another currency symbol

StevenLeighton21:
These symbols come from the currency conversion in the API, and prevent `to_f` conversions in the front end.

I am not sure if the solution is not to send them from the API (converting to currency there seems reasonable) or to strip them like this.

### DIFF
--- a/lib/hackney/income/less_dangerous_tenancy_gateway.rb
+++ b/lib/hackney/income/less_dangerous_tenancy_gateway.rb
@@ -19,7 +19,7 @@ module Hackney
         tenancies.map do |tenancy|
           t = Hackney::Income::Domain::TenancyListItem.new
           t.ref = tenancy['ref']
-          t.current_balance = tenancy['current_balance']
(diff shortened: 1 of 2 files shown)


lib/hackney/income/less_dangerous_tenancy_gateway.rb
```diff
@@ -97,7 +97,7 @@ def extract_action_diary(events:)
       def extract_agreements(agreements:)
         agreements.map do |a|
           Hackney::Income::Domain::ArrearsAgreement.new.tap do |t|
-            t.amount = a['amount']
+            t.amount = a['amount'].delete('¤').to_f
```
zuren: Should really be a test around this. Why did you choose to skip?

StevenLeighton21: There is a test, but it tests receiving an actual float rather than the string we get, I am going to update the test

PR #59, opened 2018-08-31
Show a human readable status on the list page

StevenLeighton21:
image

### DIFF
--- a/spec/lib/hackney/income/domain/tenancy_list_item_spec.rb
+++ b/spec/lib/hackney/income/domain/tenancy_list_item_spec.rb
@@ -27,5 +27,16 @@ describe Hackney::Income::Domain::TenancyListItem do
       expect(subject).to have_attributes(primary_contact_short_address: '123 Test Lane')
       expect(subject).to have_attributes(primary_contact_postcode: 'TEST')
     end
+
+    it 'should provide a human readable agreement status' do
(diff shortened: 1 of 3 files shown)


spec/lib/hackney/income/domain/tenancy_list_item_spec.rb
```diff
@@ -27,5 +27,16 @@
...
+    it 'should provide a human readable agreement status' do
+      subject.current_arrears_agreement_status = '200'
+      expect(subject.human_readable_agreement_status).to eq('Active')
+      subject.current_arrears_agreement_status = '400'
+      expect(subject.human_readable_agreement_status).to eq('Breached')
+      subject.current_arrears_agreement_status = '300'
+      expect(subject.human_readable_agreement_status).to eq('Inactive')
+      subject.current_arrears_agreement_status = nil
+      expect(subject.human_readable_agreement_status).to eq('None')
+    end
```
zuren: Think it would be better to have separate tests for each. That way, if something breaks all of them, you'll get four failing tests that you can fix at once, rather than one test which fails four times.

Could still be written quite neatly, i.e.:

```rb
context 'when 200' do
  before { subject.current_arrears_agreement_status = '200' }
  it { is_expected.to eq('Active') }
end

# etc
```

Not a necessary change if you're pressed for time, but worth considering :)

PR #62, opened 2018-09-03
Show new fields available in API data

StevenLeighton21:
Some additional fields are now available, tenure type, rent, service charges and other charges. So far, these fields are always 0.00 value on staging, but they are there.
image


madetech-jeffpinkham:
Looks good to me, just adding and displaying a few more fields from the TenancyAPI

PR #65, opened 2018-09-05
Remove RestClient

zuren:
Have seen extraordinary performance problems with this elsewhere. Not entirely sure what the cause was, but Net::HTTP is much faster.


lib/hackney/income/less_dangerous_tenancy_gateway.rb
```diff
@@ -9,12 +10,13 @@ def initialize(api_host:, api_key:)
...
-        response = RestClient.get(
-          "#{@api_host}/tenancies",
-          'x-api-key' => @api_key,
-          params: { tenancy_refs: convert_to_params_array(refs: refs) }
-        )
-        tenancies = JSON.parse(response.body)['tenancies']
+        uri = URI("#{@api_host}/tenancies?#{params_list('tenancy_refs', refs)}")
+
+        req = Net::HTTP::Get.new(uri)
+        req['X-Api-Key'] = @api_key
```
StevenLeighton21: This is case-insensitive? I've used `x-api-key` in my WIP PR, should we pick one for consistency?

zuren: Let's stick with this because it's first in? 😄

PR #70, opened 2018-09-07
Quick fix to show negative scores as zero

StevenLeighton21:


StevenLeighton21:
It's not really a bug, I think we should just be zero-ing out negatives at the API level rather than here.

PR #71, opened 2018-09-07
Show latest action

StevenLeighton21:
image

### DIFF
--- a/app/views/tenancies/index.html.erb
+++ b/app/views/tenancies/index.html.erb
@@ -9,6 +9,7 @@
       <th>Reference No.</th>
       <th>Tenant</th>
       <th>Current Agreement Status</th>
+      <th>Latest Action</th>
       <th class="numeric">Balance</th>
       <th>Priority</th>
     </tr>
@@ -31,6 +32,12 @@
         <td>
           <%= link_to(tenancy_path(id: tenancy.ref), class: 'tenancy_list__link') do %>
             <span class="tenancy_list__ref"><%= tenancy.human_readable_agreement_status %></span>
+          <% end %>
+        </td>
+        <td>
+          <%= link_to(tenancy_path(id: tenancy.ref), class: 'tenancy_list__link') do %>
+            <span class="tenancy_list__latest_action_code"><strong><%= tenancy.latest_action_code %></strong> on <br /></span>
+            <span class="tenancy_list__latest_action_date"><%= tenancy.last_action_display_date %></span>
           <% end %>
         </td>
         <td class="numeric">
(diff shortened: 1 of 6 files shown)


lib/hackney/income/domain/action_diary_entry.rb
```diff
@@ -8,6 +8,10 @@ class ActionDiaryEntry
 
         validates :balance, :code, :type, :date, :comment, :universal_housing_username,
                   presence: true
+
+        def display_date
+          Time.parse(date).to_formatted_s(:long_ordinal)
+        end
```
zuren: Wonder if this should be a helper method, in a module like TimeHelper, rather than on ActionDiaryEntry.

That way we could easily reuse it everywhere we present dates and times.

lib/hackney/income/domain/arrears_agreement.rb
```diff
@@ -10,6 +10,14 @@ class ArrearsAgreement
...
                   :start_date, :status,
                   presence: true
+
+        def start_date_display_date
+          Date.parse(start_date).to_formatted_s(:long_ordinal)
+        end
+
+        def clear_by_display_date
+          Date.parse(clear_by).to_formatted_s(:long_ordinal)
+        end
```
zuren: i.e. could remove these :D

app/views/tenancies/index.html.erb
```diff
@@ -33,6 +34,12 @@
             <span class="tenancy_list__ref"><%= tenancy.human_readable_agreement_status %></span>
           <% end %>
         </td>
+        <td>
+          <%= link_to(tenancy_path(id: tenancy.ref), class: 'tenancy_list__link') do %>
+            <span class="tenancy_list__latest_action_code"><strong><%= tenancy.latest_action_code %></strong> on <br /></span>
```
zuren: Suggest we work to get latest action code _name_ propagated through the API, so we can present that instead of a code. Would consider that _done done_ personally, as suspect most of these action codes won't be particularly meaningful.

StevenLeighton21: Agree, but showing something is better than nothing?

lib/hackney/income/domain/action_diary_entry.rb
StevenLeighton21: The only reason I haven't so far is that some of these are date only, and some are times. So I've left it with individual helpers on the model. A helper that checked date or time and potentially even took an optional arg for the format would be better, but at that point we're just wrapping the `to_formatted_s` function? I don't have a great answer.

PR #73, opened 2018-09-07
Add true options for action creation

StevenLeighton21:
This static class will allow us to add useful functions for creating our own system generated actions also.


spec/features/creating_action_diary_entry_spec.rb
```diff
@@ -37,7 +37,7 @@
       expect(page).to have_field('comment')
       expect(page).to have_field('code')
 
-      select('General Note', from: 'code')
+      select('Referred for debt advice', from: 'code')
```
madetech-jeffpinkham: I'm not quite sure I understand the syntax, but from what I can gather is this the expexted description from the code options? If so doesn't "1LS" map to "First FTA letter sent"?

StevenLeighton21: Actually seems to map to multiple values.....

madetech-jeffpinkham: That's interesting, any idea why that is the case?

StevenLeighton21: Copied directly from the spreadsheet provided, and matches that. It could be that different actions do in fact share codes?

madetech-jeffpinkham: That sounds like it might be the case, I suggest we try to verify that? Otherwise I'm happy with the PR it looks good to me 👍

StevenLeighton21: I'm going to merge for now - we can always change the mappings if the spreadsheet turns out to me inaccurate

madetech-jeffpinkham: Yep sounds good 👍

PR #76, opened 2018-09-10
Stop persisting unnecessary tenancy data on sync

zuren:
This data isn't required by the ICS, and we're persisting it for no reason. Feel this will get confusing, as we have a similar sync in the Income API for priority scores.


spec/lib/hackney/income/sql_tenancy_case_gateway_spec.rb
```diff
@@ -27,37 +15,13 @@
...
-          primary_contact_postcode: tenancy.primary_contact_postcode,
-          current_balance: tenancy.current_balance,
-          primary_contact_name: tenancy.primary_contact_name,
-          latest_action_code: tenancy.latest_action_code,
-          latest_action_date: tenancy.latest_action_date.strftime('%Y-%m-%d'),
-          current_arrears_agreement_status: tenancy.current_arrears_agreement_status,
-          score: tenancy.score,
-          band: tenancy.band
-        )
+        expect(Hackney::Models::Tenancy.exists?(ref: tenancy.ref)).to be_truthy
```
StevenLeighton21: Is there a reason to use be_truthy over eq(true)? Other than how it looks/reads?

zuren: Nope, there isn't. I generally prefer `eq(true)` as it's explicit, whereas `be_truthy` will accept 1, 'cool string', etc... But in this case it'll definitely be a bool and I felt it read nicely.

PR #78, opened 2018-09-11
Show human readable frequency and highlight active agreements

StevenLeighton21:
image

I'm using hackney-light-green because there's no precedent for this in prototypes afaik.

### DIFF
--- a/app/assets/stylesheets/components/agreement.scss
+++ b/app/assets/stylesheets/components/agreement.scss
@@ -1,5 +1,5 @@
 .agreement {
   &__active {
-    background-color: $hackney-light-green;
+    background-color: $hackney-highlight-green;
   }
 }
--- a/app/assets/stylesheets/components/agreement.scss
+++ b/app/assets/stylesheets/components/agreement.scss
@@ -1,0 +1,5 @@
+.agreement {
+  &__active {
+    background-color: $hackney-light-green;
+  }
+}
--- a/app/assets/stylesheets/components/agreement.scss
+++ b/app/assets/stylesheets/components/agreement.scss
@@ -1,5 +1,5 @@
-.agreement {
-  &__active {
+.agreement_row {
+  &--active {
     background-color: $hackney-highlight-green;
   }
 }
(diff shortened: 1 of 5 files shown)


app/views/tenancies/show.html.erb
```diff
@@ -82,11 +82,15 @@
       </thead>
       <tbody>
         <% @tenancy.agreements.each do |agreement| %>
-        <tr>
+        <tr class="agreement__<%= agreement.human_readable_agreement_status.downcase %>">
```
zuren: Clever, but kind of confusing IMO. We have to update our CSS if we update our terminology, which isn't ideal. May be a more verbose but more readable way?

Also in BEM "active", "breached", "inactive", etc would be modifiers, so would go after `--`. i.e. `agreement_row--active`

app/assets/stylesheets/components/agreement.scss
```diff
@@ -0,0 +1,5 @@
+.agreement {
+  &__active {
+    background-color: $hackney-highlight-green;
+  }
+}
```
zuren: The prototype just has alternating green/white zebra rows, rather than highlighting 'active' rows, but think I like this more. Does it still look good when we have many rows?

StevenLeighton21: That section maxes out at 5 rows in the current implementation, with no plans to change afaik.

app/views/tenancies/show.html.erb
StevenLeighton21: I'll update the BEM to be accurate, but I kind of like the mapping of statuses to visual effects. We'd also be quickly highlighting data errors such as multiple active agreements!

app/views/tenancies/show.html.erb
```diff
@@ -82,11 +82,15 @@
       </thead>
       <tbody>
         <% @tenancy.agreements.each do |agreement| %>
-        <tr>
+        <tr class="agreement_row--<%= agreement.human_readable_agreement_status.downcase %>">
```
zuren: Consider the below? I get what you're going for with the human_readable_agreement_status, but concerned it'll be troublesome.

```
class="<%= agreement.status == '200' ? 'agreement_row--active' : 'agreement_row' %>"
```

PR #79, opened 2018-09-11
Show cases assigned to the current user in the list

zuren:
The Income API has been updated (https://github.com/LBHackney-IT/lbh-income-api/pull/34) to accept a list of tenancy refs when retrieving cases. This changes the tenancy list page to request user assigned tenancy refs, rather than random ones.


spec/lib/hackney/income/list_user_assigned_cases_spec.rb
```diff
@@ -1,41 +1,37 @@
 require 'rails_helper'
 
 describe Hackney::Income::ListUserAssignedCases do
```
StevenLeighton21: Glad to see this spec back in action

PR #82, opened 2018-09-11
Add New Relic

zuren:


cob16:
@zuren what env vars do you need to set?

cob16:
sounds dandrous to default the app name to production

PR #83, opened 2018-09-11
Prevent code fetch from causing an error, just return nil

StevenLeighton21:
Probably due to receiving nil action codes or codes that are not listed. In these cases, this should now just display nil rather than causing an error.

### DIFF
--- a/lib/hackney/income/action_diary_entry_codes.rb
+++ b/lib/hackney/income/action_diary_entry_codes.rb
@@ -173,7 +173,7 @@ module Hackney
       end
 
       def self.human_readable_action_code(code)
-        all_code_options.select { |e| e.fetch(:code) == code }.first.fetch(:name)
+        all_code_options.select { |e| e.fetch(:code) == code }.first[:name]
       end
     end
   end


lib/hackney/income/action_diary_entry_codes.rb
```diff
@@ -173,7 +173,7 @@ def self.code_dropdown_options
       end
 
       def self.human_readable_action_code(code)
-        all_code_options.select { |e| e.fetch(:code) == code }.first.fetch(:name)
+        all_code_options.select { |e| e.fetch(:code) == code }.first[:name]
```
zuren: Should it display a default value instead of nil?

StevenLeighton21: Not sure there is a default value? I actually think this is caused by a pre-pivot 'scheduled action' being added to a case, and shouldn't _really_ be a concern in production

mrosel: Because this is front end,  `nil` could be valid and we allow the display layer to decide how to treat `nil`'s

StevenLeighton21:
Closing as this is no longer relevant

PR #91, opened 2018-09-17
Add register users gateway

StevenLeighton21:


spec/lib/hackney/income/register_users_gateway_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  let(:response) do
+    {
+      id: 123,
+      name: 'Richard Foster',
+      email: 'rich@richy.rich',
+      first_name: 'Richard',
+      last_name: 'Foster',
+      provider_permissions: true
+    }
```
zuren: Could do with randomisation here too, to guarantee the response isn't slimed?

lib/hackney/income/register_users_gateway.rb
```diff
@@ -0,0 +1,39 @@
+module Hackney
+  module Income
+    class RegisterUsersGateway
```
zuren: Feel like `IncomeApiUsersGateway` might be more fitting? Really easy to see what it touches that way, similar to `SqlUsersGateway`.

lib/hackney/income/register_users_gateway.rb
```diff
@@ -0,0 +1,39 @@
...
+        params = {
+          provider_uid: provider_uid,
+          provider: provider,
+          name: name,
+          email: email,
+          first_name: first_name,
+          last_name: last_name,
+          provider_permissions: provider_permissions
+        }
+        uri.query = URI.encode_www_form(params)
```
zuren: Might be quicker to read without the `params` assignment?

StevenLeighton21: How do you mean? From a human/code maintenance point of view or optimisation?

zuren: From a human perspective. Up to you really, I just find the below easier to read, as there are fewer variables. 🤷‍♀️ 

```
uri.query = URI.encode_www_form(
  provider_uid: provider_uid,
  provider: provider,
  name: name,
  email: email,
  first_name: first_name,
  last_name: last_name,
  provider_permissions: provider_permissions
)
```

PR #92, opened 2018-09-17
Correctly map frequency lookup

StevenLeighton21:
This human readable frequency was taken from the mapping spreadsheet directly, and should cover frequencies we haven't seen.

### DIFF
--- a/lib/hackney/income/domain/arrears_agreement.rb
+++ b/lib/hackney/income/domain/arrears_agreement.rb
@@ -31,8 +31,16 @@ module Hackney
         end
 
         def human_readable_frequency
-          return 'Weekly' if frequency == '1'
           return 'Monthly' if frequency == '0'
+          return 'Weekly' if frequency == '1'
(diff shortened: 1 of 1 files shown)


lib/hackney/income/domain/arrears_agreement.rb
```diff
@@ -31,8 +31,16 @@ def human_readable_breached_status
...
           return 'Monthly' if frequency == '0'
+          return 'Weekly' if frequency == '1'
+          return 'Fortnightly' if frequency == '2'
+          return '4 Weekly' if frequency == '4'
+          return '3 Monthly' if frequency == '5'
+          return '6 Monthly' if frequency == '6'
+          return 'Annually' if frequency == '7'
+          return 'Daily' if frequency == '8'
+          return 'Irregular' if frequency == '9'
+
```
zuren: A case statement might be more idiomatic. e.g.

```
case frequency
when '0' then 'Weekly'
when '1' then 'Fortnightly'
...
end
```

PR #100, opened 2018-09-23
Auth correct group

StevenLeighton21:
Auth against an email whitelist for now. There's been a lot of trouble using group based auth via Azure Active Directory, our best bet is probably going to be authing to get a user token, and using it to determine groups via something like this https://msdn.microsoft.com/en-us/library/azure/ad/graph/api/functions-and-actions#getMemberGroups.

That's quite an involved change, so this whitelist allows us to control access to the production server, while allowing access to staging using wildcards. The other option is to fall back to using only AAD to auth, but that is too 'open' to be acceptable.


app/controllers/sessions_controller.rb
```diff
@@ -39,7 +39,7 @@ def destroy
   private
 
   def user_has_ic_staff_permissions?
-    auth_hash.extra.raw_info.id_token.split('.').include?(ENV['IC_STAFF_GROUP'])
+    ENV['IC_STAFF_GROUP'].split('/').include?(auth_hash.info.email)
```
madetech-jeffpinkham: Should we be pulling from an environment variable every time? Could we potentially cache this somewhere?

PR #101, opened 2018-09-24
Actions page

StevenLeighton21:
On the show page:
image

The list page:
image


StevenLeighton21:
Note - due to the API change, this partial also renders the latest *_ten_* action diary events on the main page.

spec/controllers/tenancies_arrears_actions_controller_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'rails_helper'
+
+describe TenanciesArrearsActionsController do
+  let(:tenancy_ref) { Faker::Lorem.characters(8) }
+
+  before do
+    stub_authentication
+    stub_const('Hackney::Income::ViewActions', StubViewActions)
```
zuren: Is a stub actually required here? The `expect_any_instance_of` call should stub the ViewActions use case to return nothing.

lib/hackney/income/view_actions.rb
```diff
@@ -0,0 +1,13 @@
+module Hackney
+  module Income
+    class ViewActions
+      def initialize(actions_gateway:)
+        @actions_gateway = actions_gateway
+      end
+
+      def execute(tenancy_ref:)
+        @actions_gateway.get_actions_for(tenancy_ref: tenancy_ref)
```
zuren: Should preferably be mapping the Actions Gateway response into something the use case will consistently return. We're breaking the boundary between controller/gateway here, and can't rely on the use case output.

config/routes.rb
```diff
@@ -9,6 +9,7 @@
   get '/tenancies/:id/transactions', to: 'tenancies_transactions#index', as: :tenancies_transactions
   get '/tenancies/:id/action_diary', to: 'action_diary_entry#show', as: :action_diary_entry
   post '/tenancies/:id/action_diary', to: 'action_diary_entry#create', as: :create_action_diary_entry
+  get '/tenancies/:id/actions', to: 'tenancies_arrears_actions#index', as: :tenancies_actions
```
zuren: Looks good, but highlights the weirdness of `action_diary_entry#show` instead of `action_diary_entry#new`. These should probably be grouped together as a single resource.

i.e.

`GET /tenancies/:id/actions` -> `tenancies_arrears_actions#index`
`GET /tenancies/:id/actions/new` -> `action_diary_entry#new`
`POST /tenancies/:id/actions` -> `action_diary_entry#create`

Don't know if this needs to change, but it's tech debt.

app/views/tenancies_arrears_actions/index.html.erb
```diff
@@ -0,0 +1,16 @@
+<div class="grid-row">
```
zuren: No tenancy header on this page? Will look different to our other pages.

lib/hackney/income/view_actions.rb
StevenLeighton21: I consciously did this to remain consistent with the rest of the gateway/use cases. Begin changing as we go or remain consistent?

config/routes.rb
StevenLeighton21: Agree, I thought it was odd when I was putting it together, and wondered if the `tenancies_actions` controller ought to be considered moot. I came down to the decision based on context, this is actions for a tenancy, vs _new_ arrears actions in a controller similar to the sms/email controllers.

app/views/tenancies_arrears_actions/index.html.erb
StevenLeighton21: I think the header partial is kinda superfluous and it adds a layer of complexity and negative performance to retrieve the whole tenancy, which is required to render that partial. I would personally lean towards removing it from the other pages, but I'm happy to put it in if that's preferred.

spec/controllers/tenancies_arrears_actions_controller_spec.rb
StevenLeighton21: I do love rspec.

lib/hackney/income/view_actions.rb
zuren: Would say we should be changing as we go - the original use cases all did this, it's fairly recent that they haven't been.

config/routes.rb
zuren: Would still say that _resourcefully_ they're all related to a single tenancy, so would make sense to have them routed accordingly.

app/views/tenancies_arrears_actions/index.html.erb
zuren: I can see that, would probably opt for whatever's most performant.

StevenLeighton21:
Updated with review comments - I've check the new routes work for show/create/index on arrears actions 👍 
There's other areas where would be improving the domain object -> view data barrier as we go.

PR #102, opened 2018-09-26
Add display date to arrears action hash

StevenLeighton21:
image

A recent change meant the front was trying to display a hash rather than the domain object directly, which did not include a formatted display date. Rectified this.


StevenLeighton21:
It is equivalent @madetech-jeffpinkham - Those scheduled events contain an actual timestamp (they're not even used currently and likely to be removed) which would throw off the `display_date` conversion. As a string, it's able to parse it and give a formatted date.

madetech-jeffpinkham:
Yep sweet 👍

PR #103, opened 2018-09-26
Contact details for tenancy

StevenLeighton21:
image
Future PR - When sending SMS/email select from available contacts listed on the tenancy
Future PR - Improve styling where more than one contact is present (it'll add more lists below the first, currently)

I decided to simple skip the step and return as if there are no contact details in the staging environment, since no communications are sent to real numbers via staging, and names and addresses are already obfuscated for the tenancy itself.

### DIFF
--- a/app/views/common/_contact_details.html.erb
+++ b/app/views/common/_contact_details.html.erb
@@ -1,0 +1,11 @@
+<ul class="list">
+  <li>Title: <%= contact[:title] %></li>
+  <li>Full Name: <%= contact[:full_name] %></li>
+  <li>Mobile Phone: <%= contact[:telephone_2] %></li>
+  <li>Home Phone: <%= contact[:telephone_1] %></li>
+  <li>E-Mail: <%= contact[:email_address] %></li>
+  <li>
+    <%= link_to 'Send SMS', tenancy_sms_path(id: @tenancy.ref), class:'button' %>
+    <%= link_to 'Send Email', tenancy_email_path(id: @tenancy.ref), class:'button' %>
+  </li>
+</ul>
--- a/app/views/common/_contact_details.html.erb
+++ b/app/views/common/_contact_details.html.erb
@@ -1,11 +1,9 @@
 <ul class="list">
   <li>Title: <%= contact[:title] %></li>
-  <li>Full Name: <%= contact[:full_name] %></li>
-  <li>Mobile Phone: <%= contact[:telephone_2] %></li>
+  <li>First Name: <%= contact[:first_name] %></li>
+  <li>Last Name: <%= contact[:last_name] %></li>
   <li>Home Phone: <%= contact[:telephone_1] %></li>
+  <li>Mobile Phone: <%= contact[:telephone_2] %></li>
+  <li>Other Phone: <%= contact[:telephone_3] %></li>
   <li>E-Mail: <%= contact[:email_address] %></li>
-  <li>
-    <%= link_to 'Send SMS', tenancy_sms_path(id: @tenancy.ref), class:'button' %>
-    <%= link_to 'Send Email', tenancy_email_path(id: @tenancy.ref), class:'button' %>
-  </li>
 </ul>
--- a/app/views/tenancies/show.html.erb
+++ b/app/views/tenancies/show.html.erb
@@ -38,11 +38,21 @@
     <h2>Personal details</h2>
     <div class="grid-row">
       <div class="column-half">
-        <h3>Current address</h3>
+        <h3>Tenancy address</h3>
         <ul class="list">
           <li><%= @tenancy.primary_contact_long_address %></li>
           <li><%= @tenancy.primary_contact_postcode %></li>
         </ul>
+        <% if !@tenancy.contacts[0].nil? %>
+          <h3>Primary Tenant</h3>
+          <ul class="list">
+            <li><%= @tenancy.contacts[0][:full_name] %></li>
+            <li><%= @tenancy.contacts[0][:address_line_1] %></li>
+            <li><%= @tenancy.contacts[0][:address_line_2] %></li>
+            <li><%= @tenancy.contacts[0][:address_line_3] %></li>
+            <li><%= @tenancy.contacts[0][:post_code] %></li>
+          </ul>
+        <% end %>
       </div>
       <div class="column-half">
         <h3>Contact details</h3>
@@ -52,6 +62,8 @@
           <% @tenancy.contacts.each do |contact| %>
           <%= render('common/contact_details', contact: contact) %>
           <% end %>
+          <%= link_to 'Send SMS', tenancy_sms_path(id: @tenancy.ref), class:'button' %>
+          <%= link_to 'Send Email', tenancy_email_path(id: @tenancy.ref), class:'button' %>
         <% end %>
       </div>
     </div>
--- a/app/views/tenancies/show.html.erb
+++ b/app/views/tenancies/show.html.erb
@@ -44,17 +44,15 @@
           <li><%= @tenancy.primary_contact_postcode %></li>
         </ul>
       </div>
-      <div class="column-half fake">
+      <div class="column-half">
         <h3>Contact details</h3>
-        <ul class="list">
-          <li>Mobile Phone: <%= @tenancy.primary_contact_phone %></li>
-          <li>Home Phone: <%= @tenancy.primary_contact_phone %></li>
-          <li>E-Mail: <%= @tenancy.primary_contact_email %></li>
-          <li>
-            <%= link_to 'Send SMS', tenancy_sms_path(id: @tenancy.ref), class:'button' %>
-            <%= link_to 'Send Email', tenancy_email_path(id: @tenancy.ref), class:'button' %>
-          </li>
-        </ul>
+        <% if @tenancy.contacts.empty? %>
+        Contact data appears to be missing for this tenant.
+        <% else %>
+          <% @tenancy.contacts.each do |contact| %>
+          <%= render('common/contact_details', contact: contact) %>
+          <% end %>
+        <% end %>
       </div>
     </div>
   </div>
--- a/lib/hackney/income/less_dangerous_tenancy_gateway.rb
+++ b/lib/hackney/income/less_dangerous_tenancy_gateway.rb
@@ -126,7 +126,7 @@ module Hackney
             t.telephone_2 = c['telephone2']
             t.telephone_3 = c['telephone3']
             t.cautionary_alert = c['cautionary_alert']
-            t.property_cautionary_alert = c['propertyCautionary_alert']
+            t.property_cautionary_alert = c['property_cautionary_alert']
             t.house_ref = c['house_ref']
             t.title = c['title']
             t.full_address_display = c['full_address_display']
--- a/lib/hackney/income/less_dangerous_tenancy_gateway.rb
+++ b/lib/hackney/income/less_dangerous_tenancy_gateway.rb
@@ -108,32 +108,32 @@ module Hackney
 
         contacts = JSON.parse(res.body)['data']['contacts']
 
-        return [] if contacts.blank?
+        return [] if contacts.blank? || Rails.env.staging?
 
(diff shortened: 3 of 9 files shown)


app/views/tenancies/show.html.erb
```diff
@@ -38,23 +38,33 @@
...
-          <li>
-            <%= link_to 'Send SMS', tenancy_sms_path(id: @tenancy.ref), class:'button' %>
-            <%= link_to 'Send Email', tenancy_email_path(id: @tenancy.ref), class:'button' %>
-          </li>
-        </ul>
+        <% if @tenancy.contacts.empty? %>
+        Contact data appears to be missing for this tenant.
+        <% else %>
+          <% @tenancy.contacts.each do |contact| %>
+          <%= render('common/contact_details', contact: contact) %>
```
madetech-jeffpinkham: nice, like this 👍

madetech-jeffpinkham:
Merge at your leisure sir

app/views/tenancies/show.html.erb
```diff
@@ -38,23 +38,33 @@
...
           <li><%= @tenancy.primary_contact_postcode %></li>
         </ul>
+        <% if !@tenancy.contacts[0].nil? %>
+          <h3>Primary Tenant</h3>
+          <ul class="list">
+            <li><%= @tenancy.contacts[0][:full_name] %></li>
+            <li><%= @tenancy.contacts[0][:address_line_1] %></li>
+            <li><%= @tenancy.contacts[0][:address_line_2] %></li>
+            <li><%= @tenancy.contacts[0][:address_line_3] %></li>
+            <li><%= @tenancy.contacts[0][:post_code] %></li>
```
madetech-jeffpinkham: Do we want email address here?

lib/hackney/income/less_dangerous_tenancy_gateway.rb
```diff
@@ -98,6 +98,46 @@ def get_tenancy(tenancy_ref:)
...
+            t.address_line_3 = c['address_line3']
+            t.first_name = c['first_name']
+            t.last_name = c['last_name']
+            t.full_name = c['full_name']
+            t.larn = c['larn']
+            t.telephone_1 = c['telephone1']
+            t.telephone_2 = c['telephone2']
+            t.telephone_3 = c['telephone3']
+            t.cautionary_alert = c['cautionary_alert']
+            t.property_cautionary_alert = c['propertyCautionary_alert']
```
madetech-jeffpinkham: should this be c['property_cautionary_alert']

madetech-jeffpinkham: Your mapping is right I've not outputted it in snake case

madetech-jeffpinkham: @StevenLeighton21  I've created a task for me to change this. If you want to tackle this now we can

app/views/tenancies/show.html.erb
StevenLeighton21: Post code is correct I think, this line is the 'primary tenant address' (because it can be different from the address of the property)

lib/hackney/income/less_dangerous_tenancy_gateway.rb
StevenLeighton21: It's not urgent - as long as we update this format with the one from the API after the change. I'd say add a task for both and put them in the backlog.

app/views/tenancies/show.html.erb
madetech-jeffpinkham: Ohh ok

lib/hackney/income/less_dangerous_tenancy_gateway.rb
madetech-jeffpinkham: https://github.com/LBHackney-IT/LBHTenancyAPI/pull/63

madetech-jeffpinkham: It's up to you when you want to integrate those changes.

madetech-jeffpinkham: Although it might be easier to do them now?

StevenLeighton21: Agree, now that PR is in I can add it here.

lib/hackney/income/less_dangerous_tenancy_gateway.rb
```diff
@@ -126,7 +126,7 @@ def get_contacts_for(tenancy_ref:)
             t.telephone_2 = c['telephone2']
             t.telephone_3 = c['telephone3']
             t.cautionary_alert = c['cautionary_alert']
-            t.property_cautionary_alert = c['propertyCautionary_alert']
+            t.property_cautionary_alert = c['property_cautionary_alert']
```
madetech-jeffpinkham: I've merged the PR regarding this

StevenLeighton21: And the field is now updated :)

madetech-jeffpinkham: sweet 👍

app/views/common/_contact_details.html.erb
```diff
@@ -0,0 +1,9 @@
+<ul class="list">
+  <li>Title: <%= contact[:title] %></li>
+  <li>First Name: <%= contact[:first_name] %></li>
+  <li>Last Name: <%= contact[:last_name] %></li>
```
zuren: Is this preferable to `<%= contact[:title] %> <%= contact[:first_name] %> <%= contact[:last_name] %>`

app/views/tenancies/show.html.erb
zuren: Is this assuming that the first tenancy is the responsible one?

In the Manage a Tenancy APIs, primary contacts were distinguished by a "responsible" key I think. Is our implementation similar?

app/views/tenancies/show.html.erb
zuren: Would it be worth noting how a user would/should handle this? If they needed contact information and it wasn't displayed here? What would they want to know?

lib/hackney/income/less_dangerous_tenancy_gateway.rb
```diff
@@ -98,6 +98,46 @@ def get_tenancy(tenancy_ref:)
...
+        uri = URI("#{@api_host}/tenancies/#{ERB::Util.url_encode(tenancy_ref)}/contacts")
+
+        req = Net::HTTP::Get.new(uri)
+        req['X-Api-Key'] = @api_key
+
+        res = Net::HTTP.start(uri.hostname, uri.port, use_ssl: true) { |http| http.request(req) }
+
+        contacts = JSON.parse(res.body)['data']['contacts']
+
+        return [] if contacts.blank? || Rails.env.staging?
```
zuren: So no contacts in staging? How come?

lib/hackney/income/stub_tenancy_gateway_builder.rb
```diff
@@ -105,6 +138,7 @@ def create_tenancy(attributes)
                 t.primary_contact_email = 'test@example.com'
                 t.arrears_actions = [action]
                 t.agreements = [agreement]
+                t.contacts = nil
```
zuren: Should this be an empty array?

spec/lib/hackney/income/less_dangerous_tenancy_gateway_spec.rb
```diff
@@ -262,6 +262,92 @@
...
+    before do
+      stub_request(:get, 'https://example.com/api/tenancies/FAKE%2F01/contacts')
+        .to_return(body: stub_single_response.to_json)
+      stub_request(:get, 'https://example.com/api/tenancies/FAKE%2F02/contacts')
+        .to_return(body: stub_joint_response.to_json)
+      stub_request(:get, 'https://example.com/api/tenancies/FAKE%2F03/contacts')
+        .to_return(body: stub_empty_response_1.to_json)
+      stub_request(:get, 'https://example.com/api/tenancies/FAKE%2F04/contacts')
+        .to_return(body: stub_empty_response_2.to_json)
+    end
```
zuren: I feel this test would be more readable if we put the appropriate `let` blocks and `before` statement containing a `stub_request` call for each `context`, rather than setting them all up for every test and distinguishing by tenancy ref inside out `context`s.

i.e. each context would look like:

```
context  '...' do
  let(:stub_response) { generate_contacts_response(...) }
  before do
    stub_request(:get, 'https://example.com/api/tenancies/FAKE%2F01/contacts')
        .to_return(body: stub_response.to_json)
  end

  it '...' do
  end
end
```

spec/lib/hackney/income/view_tenancy_spec.rb
```diff
@@ -35,6 +35,32 @@
...
+        expect(subject.contacts[0].fetch(:property_cautionary_alert)).to eq(false)
+        expect(subject.contacts[0].fetch(:house_ref)).to eq('98765')
+        expect(subject.contacts[0].fetch(:title)).to eq('Mr.')
+        expect(subject.contacts[0].fetch(:full_address_display)).to eq('123 Test Road, Delivery City')
+        expect(subject.contacts[0].fetch(:full_address_search)).to eq('Search')
+        expect(subject.contacts[0].fetch(:post_code)).to eq('E0 123')
+        expect(subject.contacts[0].fetch(:date_of_birth)).to eq('12th March, 1976')
+        expect(subject.contacts[0].fetch(:hackney_homes_id)).to eq('1209')
+      end
+
```
zuren: Unfortunately no proof that this isn't slimed. I understand why this wasn't done, but it's reducing our confidence in our suite. Is our organisation making it too difficult to stub easily?

app/views/common/_contact_details.html.erb
StevenLeighton21: I would say so, yes. My reasoning is that as a user, when you go to send a message to this contact, you'll be populating templates with various combos of first name, last name, title and _full name_ which are all unique fields returned from the data source. Image someone with multiple middle names - you'd have a given name, surname and a true full name visible this way.

app/views/tenancies/show.html.erb
StevenLeighton21: Huge assumption, yes. AFAIK the responsible contact is a separate API.

app/views/common/_contact_details.html.erb
zuren: Makes sense 👍

app/views/tenancies/show.html.erb
zuren: @madetech-jeffpinkham @LBHRShetty have we not considered this in our API?

lib/hackney/income/less_dangerous_tenancy_gateway.rb
zuren: Makes sense. Making it difficult to verify in our release workflow, but it's alright.

Missed it in the description, sorry!

lib/hackney/income/stub_tenancy_gateway_builder.rb
zuren: Right, sweet!

spec/lib/hackney/income/view_tenancy_spec.rb
zuren: It is tedious. It's a balance between readability and confidence.

If there's a way to reduce boilerplate to keep confidence high, and tedium low, we should be seeking it wherever poss. Understand the reasoning though.

lib/hackney/income/less_dangerous_tenancy_gateway.rb
StevenLeighton21: I'm happy verifying that none of them ever show up in staging, personally?

spec/lib/hackney/income/view_tenancy_spec.rb
StevenLeighton21: Agree, I don't _like_ it, I feel the gain vs time investment isn't worth it right now.

PR #104, opened 2018-09-26
Also add a display date for actions via the all actions endpoint

StevenLeighton21:
image

This was resolved for tenancies#show, but not for actions#index, as the actions come from a different endpoint via a different use case, and not resolving this in the previous PR was an oversight that became obvious when testing on Staging.

### DIFF
--- a/spec/lib/hackney/income/view_actions_spec.rb
+++ b/spec/lib/hackney/income/view_actions_spec.rb
@@ -8,7 +8,7 @@ describe Hackney::Income::ViewActions do
         t.balance = Faker::Number.decimal(2)
         t.code = Faker::Lorem.characters(3)
         t.type = Faker::Lorem.characters(3)
-        t.date = Faker::Date.forward(100)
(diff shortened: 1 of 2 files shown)


spec/lib/hackney/income/view_actions_spec.rb
```diff
@@ -37,6 +38,7 @@
         code: actions[1].code,
         type: actions[1].type,
         date: actions[1].date,
+        display_date: actions[1].display_date,
```
madetech-jeffpinkham: Is there a reason why we are manually referencing actions in an array?

StevenLeighton21: Yeah, the test creates two mock domain objects with randomised values, and returns that from the mock gateway. The hash you commented on is constructing an expected output from the use case, and then a comparison is made. 
The test is really very very close to the implementation, but that is ok for me as the use case is mapping domain objects into view objects.

PR #105, opened 2018-10-01
Transaction summary

StevenLeighton21:
oct-01-2018 17-49-58

image

*Why*
Users would like to see payments grouped by date, with a total, and the ability to click the row to see the breakdown of charges for that day, in order to condense the payment list (which can be 1000s of entries).

*What*
Currently, we get a list of transactions for an account that looks something like {type, amount debited/credited, date} and we get multiple small outgoings each day that _rent_ is debited.  Now this list is grouped by incoming/outgoing for a particular date, and the table shows a row for this summary with the total. Clicking the row expands it to show the individual transactions for that date. Clicking again will collapse back to just the summary row.

### DIFF
--- a/app/views/common/_tenancy_transactions_list.html.erb
+++ b/app/views/common/_tenancy_transactions_list.html.erb
@@ -1,4 +1,4 @@
-<table>
+<table class="transactions-table">
   <thead>
     <th>Date</th>
     <th>Transaction</th>
@@ -7,7 +7,7 @@
     <th class="numeric">Balance</th>
   </thead>
   <% transactions.each do |transaction_summary| %>
-  <tbody class="tenancy_payments_summary">
+  <tbody class="transactions-table__group--summary">
     <tr>
       <td><%= transaction_summary.fetch(:date).to_formatted_s(:long_ordinal) %></td>
       <td><%= transaction_summary.fetch(:description) %></td>
@@ -20,7 +20,7 @@
       </td>
     </tr>
   </tbody>
-    <tbody class="tenancy_payments_list">
+    <tbody class="transactions-table__group--breakdown">
       <% transaction_summary[:transactions].each do |transaction| %>
       <tr>
         <td><%= transaction_summary.fetch(:date).to_formatted_s(:long_ordinal) %></td>
@@ -38,7 +38,7 @@
 </table>
 
 <script>
-$('.tenancy_payments_summary').click(function () {
+$('.transactions-table__group--summary').click(function () {
   $(this).closest('tbody').next().toggle()
 })
 </script>
--- a/app/views/common/_tenancy_transactions_list.html.erb
+++ b/app/views/common/_tenancy_transactions_list.html.erb
@@ -1,10 +1,8 @@
 <table>
   <thead>
     <th>Date</th>
-    <th>Total Value</th>
     <th>Transaction</th>
     <th>Code</th>
-    <th>Date</th>
     <th class="numeric">Value</th>
     <th class="numeric">Balance</th>
   </thead>
@@ -12,28 +10,22 @@
   <tbody class="tenancy_payments_summary">
     <tr>
       <td><%= transaction_summary.fetch(:date).to_formatted_s(:long_ordinal) %></td>
-      <td>
+      <td><%= transaction_summary.fetch(:description) %></td>
+      <td></td>
+      <td class="numeric <%= transaction_summary.fetch(:total_charge) > 0 ? 'negative' : 'positive' %>">
         <%= transaction_summary.fetch(:total_charge) > 0 ? '£%.2f' % transaction_summary.fetch(:total_charge) : '-£%.2f' % transaction_summary.fetch(:total_charge).abs %>
       </td>
-      <td></td>
-      <td></td>
-      <td></td>
-      <td></td>
-      <td></td>
+      <td class="numeric">
+        £<%= '%.2f' % transaction_summary.fetch(:final_balance) %>
+      </td>
     </tr>
   </tbody>
     <tbody class="tenancy_payments_list">
       <% transaction_summary[:transactions].each do |transaction| %>
       <tr>
         <td><%= transaction_summary.fetch(:date).to_formatted_s(:long_ordinal) %></td>
-        <td>
-          <%= transaction_summary.fetch(:total_charge) > 0 ? '£%.2f' % transaction_summary.fetch(:total_charge) : '-£%.2f' % transaction_summary.fetch(:total_charge).abs %>
-        </td>
         <td><%= transaction.fetch(:description) %></td>
         <td><%= transaction.fetch(:type) %></td>
-        <td title="<%= transaction.fetch(:timestamp) %>">
-          <%= transaction.fetch(:timestamp).to_formatted_s(:long_ordinal) %>
-        </td>
         <td class="numeric <%= transaction.fetch(:value) > 0 ? 'negative' : 'positive' %>">
           <%= transaction.fetch(:value) > 0 ? '£%.2f' % transaction.fetch(:value) : '-£%.2f' % transaction.fetch(:value).abs %>
         </td>
--- a/app/views/common/_tenancy_transactions_list.html.erb
+++ b/app/views/common/_tenancy_transactions_list.html.erb
@@ -13,7 +13,8 @@
       <td>
         <%= transaction_summary.fetch(:description) %>
         <% if transaction_summary.fetch(:transactions).count > 1 %>
-        <span><br/>+ click to expand</span
+        <span class="transactions-table__expand-link"><br/>+ click to expand</span>
+        <span class="transactions-table__collapse-link transactions-table__collapse-link--hidden"><br/>- click to collapse</span>
         <% end %>
       </td>
       <td></td>
@@ -45,5 +46,7 @@
 <script>
 $('.transactions-table__group--summary').click(function () {
   $(this).closest('tbody').next().toggleClass('transactions-table__group--hidden')
+  $(this).find('.transactions-table__expand-link').toggleClass('transactions-table__expand-link--hidden')
+  $(this).find('.transactions-table__collapse-link').toggleClass('transactions-table__collapse-link--hidden')
 })
 </script>
--- a/app/views/common/_tenancy_transactions_list.html.erb
+++ b/app/views/common/_tenancy_transactions_list.html.erb
@@ -1,59 +1,34 @@
-
-    <% transactions.each do |transaction_summary| %>
-    <table>
-      <thead>
-        <th>Date</th>
-        <th class="numeric">Total Value</th>
-      </thead>
-      <tbody>
-    <tr>
-      <td><%= transaction_summary.fetch(:date) %></td>
-      <%= transaction_summary.fetch(:total_charge) > 0 ? '£%.2f' % transaction_summary.fetch(:total_charge) : '-£%.2f' % transaction_summary.fetch(:total_charge).abs %>
-      <td>
-        <table>
-          <thead>
-            <th>Transaction</th>
-            <th>Code</th>
-            <th>Date</th>
-            <th class="numeric">Value</th>
-            <th class="numeric">Balance</th>
-          </thead>
-          <tbody>
-              <% transaction_summary[:transactions].each do |transaction| %>
-              <tr>
-                <td><%= transaction.fetch(:description) %></td>
-                <td><%= transaction.fetch(:type) %></td>
-                <td title="<%= transaction.fetch(:timestamp) %>">
-                  <%= transaction.fetch(:timestamp).strftime('%B %e %Y') %>
-                </td>
-                <td class="numeric <%= transaction.fetch(:value) > 0 ? 'negative' : 'positive' %>">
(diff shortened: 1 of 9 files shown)


lib/hackney/income/view_tenancy.rb
```diff
@@ -101,6 +103,31 @@ def calculate_final_balance(next_transaction, current_balance)
...
+
+        outgoing.group_by { |d| d[:timestamp] }.each do |date, t|
+          summarised_transactions << { date: date.to_date.to_formatted_s(:long_ordinal), total_charge: charges(t), transactions: t }
+        end
+
+        incoming.group_by { |d| d[:timestamp] }.each do |date, t|
+          summarised_transactions << { date: date.to_date.to_formatted_s(:long_ordinal), total_charge: charges(t), transactions: t }
+        end
+
+        summarised_transactions.sort_by { |summary| summary[:date] }.reverse
```
emileswarts:
```ruby
 sorted_transactions = transactions.sort_by  { |t| t.fetch(:value) }

 formatted_transactions = sorted_transactions.map do |t|
  {
    date: date.to_date.to_formatted_s(:long_ordinal),
    total_charge: charges(t),
    transactions: t 
  }
 end

formatted_transactions.sort_by { |summary| summary[:date] }.reverse
```

Haven't tested this, but think it's equivalent (but less imperative)?

emileswarts: Also is the final sort going to sort by string date?
Wonder if there is value in keeping a date object around until you get to displaying it.
Have the view cast it to string.

spec/lib/hackney/income/view_tenancy_spec.rb
```diff
@@ -63,27 +63,31 @@
...
         expect(subject.transactions).to include(
-          id: '123-456-789',
-          timestamp: Time.new(2018, 1, 1, 0, 0, 0),
-          tenancy_ref: '3456789',
-          description: 'Rent Payment',
-          value: -50.00,
-          type: 'RPY',
-          final_balance: 1200.99
+          date: 'January 1st, 2018',
+          total_charge: -50.0,
```
emileswarts: Feels like total_charge implies negative without having to change the actual value to negative

app/views/tenancies/show.html.erb
```diff
@@ -188,10 +188,10 @@
     var ctx = document.getElementById('balance_chart')
     var transactions = <%= @tenancy.transactions.map do |t|
       {
-        description: t.fetch(:description),
-        date: t.fetch(:timestamp).to_date.to_s,
-        displayValue: t.fetch(:value) >= 0 ? '£%.2f' % t.fetch(:value) : '-£%.2f' % t.fetch(:value).abs,
-        finalBalance: t.fetch(:final_balance).round(2)
+        #description: //t.fetch(:description),
```
adrianclay: Commented code?

lib/hackney/income/view_tenancy.rb
StevenLeighton21: I was wrong about the dates, have improved the sorting and displaying, good catch!

app/views/common/_tenancy_transactions_list.html.erb
```diff
@@ -1,26 +1,52 @@
...
-        £<%= '%.2f' % transaction.fetch(:final_balance) %>
+      <td><%= transaction_summary.fetch(:date).to_formatted_s(:long_ordinal) %></td>
+      <td>
+        <%= transaction_summary.fetch(:total_charge) > 0 ? '£%.2f' % transaction_summary.fetch(:total_charge) : '-£%.2f' % transaction_summary.fetch(:total_charge).abs %>
       </td>
+      <td></td>
+      <td></td>
+      <td></td>
+      <td></td>
+      <td></td>
```
zuren: Feel we should fill these with the right computed values so the 'sum' row sits alongside normal rows in a readable way. e.g.

image

app/views/common/_tenancy_transactions_list.html.erb
```diff
@@ -1,26 +1,52 @@
...
     <th>Transaction</th>
     <th>Code</th>
     <th>Date</th>
     <th class="numeric">Value</th>
     <th class="numeric">Balance</th>
   </thead>
-  <tbody>
-    <% transactions.each do |transaction| %>
+  <% transactions.each do |transaction_summary| %>
+  <tbody class="tenancy_payments_summary">
```
zuren: Need to remember BEM conventions whenever defining a new class. This should take into account the block (i.e. whole table), element (i.e. tbody) and modifer (i.e. summary). Then we'd have logically grouped component classes like:

```
.transactions-table
.transactions-table__group
.transactions-table__group--summary
.transactions-table__group--breakdown
.transactions-table__group--hidden
```

app/views/common/_tenancy_transactions_list.html.erb
```diff
@@ -1,26 +1,52 @@
...
-        <%= transaction.fetch(:timestamp).strftime('%B %e %Y') %>
-      </td>
-      <td class="numeric <%= transaction.fetch(:value) > 0 ? 'negative' : 'positive' %>">
-        <%= transaction.fetch(:value) > 0 ? '£%.2f' % transaction.fetch(:value) : '-£%.2f' % transaction.fetch(:value).abs %>
-      </td>
-      <td class="numeric">
-        £<%= '%.2f' % transaction.fetch(:final_balance) %>
+      <td><%= transaction_summary.fetch(:date).to_formatted_s(:long_ordinal) %></td>
+      <td>
+        <%= transaction_summary.fetch(:total_charge) > 0 ? '£%.2f' % transaction_summary.fetch(:total_charge) : '-£%.2f' % transaction_summary.fetch(:total_charge).abs %>
```
zuren: Suggest we take the opportunity to move this into a helper, if it isn't already. `display_amount(charge)`?

spec/lib/hackney/income/view_tenancy_spec.rb
madetech-jeffpinkham: Is it possible to group the positives and negatives by date? So we can show them how much rent got charged on that day and how much rent got paid on that day?

app/views/common/_tenancy_transactions_list.html.erb
```diff
@@ -1,26 +1,52 @@
...
+          £<%= '%.2f' % transaction.fetch(:final_balance) %>
+        </td>
+      </tr>
+      <% end %>
+  <% end %>
 </table>
+
+<script>
+$('.tenancy_payments_summary').click(function () {
+  $(this).closest('tbody').next().toggle()
```
zuren: Recommend toggling a class like `--breakdown` using `.toggleClass(class)` instead of relying on `.toggle()` changing our CSS on an element basis.

app/views/tenancies/show.html.erb
```diff
@@ -186,12 +186,12 @@
...
+    var transactions = <%= from_last_year(@tenancy.transactions).map do |t|
       {
-        description: t.fetch(:description),
-        date: t.fetch(:timestamp).to_date.to_s,
-        displayValue: t.fetch(:value) >= 0 ? '£%.2f' % t.fetch(:value) : '-£%.2f' % t.fetch(:value).abs,
-        finalBalance: t.fetch(:final_balance).round(2)
+        description: 'Summary for -',
+        date: t.fetch(:date),
+        displayValue: t.fetch(:total_charge) >= 0 ? '£%.2f' % t.fetch(:total_charge) : '-£%.2f' % t.fetch(:total_charge).abs,
+        finalBalance: t.fetch(:transactions).first.fetch(:final_balance)
```
zuren: Do we not need the rounding anymore?

lib/hackney/income/view_tenancy.rb
```diff
@@ -101,6 +103,23 @@ def calculate_final_balance(next_transaction, current_balance)
...
+        incoming = transactions.partition { |v| v[:value].negative? }.first
+        outgoing = transactions.partition { |v| v[:value].positive? }.first
+
+        outgoing.group_by { |d| d[:timestamp] }.each do |date, t|
+          summarised_transactions << { date: date, total_charge: t.sum { |c| c.fetch(:value) }, transactions: t }
+        end
+
+        incoming.group_by { |d| d[:timestamp] }.each do |date, t|
+          summarised_transactions << { date: date, total_charge: t.sum { |c| c.fetch(:value) }, transactions: t }
+        end
```
zuren: I don't really see the value in grouping `incoming` transactions?

lib/hackney/income/view_tenancy.rb
```diff
@@ -101,6 +103,23 @@ def calculate_final_balance(next_transaction, current_balance)
...
+
+        summarised_transactions.sort_by { |summary| summary[:date] }.reverse
```
zuren: This behaviour means a set of transactions like these:

HOUSING BENEFIT: -£50
RENT: +£100
WATER: +£5
PAYPOINT PAYMENT: -£55

would display out of order. Feel like I might want to see these in order as a case worker?

zuren: Think I'm right about this? Fairly hard to read. Would the order be like the below?

[RENT, WATER] then [HOUSING BENEFIT, PAYPOINT PAYMENT]

app/views/common/_tenancy_transactions_list.html.erb
```diff
@@ -1,26 +1,52 @@
...
+})
+</script>
```
zuren: I'd also recommend moving this code into a separate JS file, similar to BalanceGraph.js.

zuren: Also - seems like this is currently hiding the summary row itself after being toggled off, from looking at your screenshot?

app/views/tenancies/show.html.erb
StevenLeighton21: We sure do, if we want to display it!

lib/hackney/income/view_tenancy.rb
StevenLeighton21: What if you have incoming and outgoing on the same day? Grouped together turned out to be much less helpful on the list page

lib/hackney/income/view_tenancy.rb
StevenLeighton21: No? 
image

app/views/common/_tenancy_transactions_list.html.erb
StevenLeighton21: It did but since resolved in a later commit.

spec/lib/hackney/income/view_tenancy_spec.rb
StevenLeighton21: That's what it does :)

madetech-jeffpinkham: ahhh, ignore me

app/views/common/_tenancy_transactions_list.html.erb
StevenLeighton21: Agree on the move but I'd like to wrap this and will look to merge with the script tag where it is unless you feel strongly about moving it 👍

lib/hackney/income/view_tenancy.rb
zuren: I imagined you could have something like:

- incoming
- incoming
- outgoing (group)
- incoming

lib/hackney/income/view_tenancy.rb
zuren: Those aren't on the same day, though. What would that behaviour be?

app/views/common/_tenancy_transactions_list.html.erb
zuren: Fine to leave it for now.

zuren:
Looks good man. Remaining comments are:

- why did we move the date in front of the transaction type name?
- think we could do a better job of distinguishing which rows are expandable. maybe a very visible + mark, or a different colour on the "click for breakdown" text?
- think we could do a better job of distinguishing which rows are in the breakdown. Maybe hide the date?
- also, now that we're using a background other than white, the lack of padding on the table rows is more evident. think we should add some left and right padding to them.

lib/hackney/income/view_tenancy.rb
StevenLeighton21: You just have another row of the same date with an entirely positive summary. Merging the two into one summary might look nice to us but I feel it's entirely unhelpful for a user.

lib/hackney/income/view_tenancy.rb
StevenLeighton21: That's what we have? Just sorted by date

app/views/common/_tenancy_transactions_list.html.erb
StevenLeighton21: See updated description

StevenLeighton21:
> why did we move the date in front of the transaction type name?
It's now grouped into summaries, by date. It's the first thing you'll look for?

StevenLeighton21:
I've also added a change to the graph that only displays the last year of summaries, looks way less cluttered.

zuren:
Looks good 👍 Keen to see the updated graph.

Final comment is on the "+ click to expand" text. It should change to something like "- click to hide breakdown" if it's already expanded. Could wrap the text in a class like "transactions-table__expand-link", select that with jQuery and amend the content.

StevenLeighton21:
@cob16 Pointed out a browser test around this accordion would be advisable. Agree, but deferring.

PR #106, opened 2018-10-03
Retrieve and use responsible field to highlight responsible tenants

StevenLeighton21:
image

TODO:
Improve UX for cases where contacts list size > 2

### DIFF
--- a/app/views/common/_contact_details.html.erb
+++ b/app/views/common/_contact_details.html.erb
@@ -1,4 +1,7 @@
-<ul class="list">
+<ul class="list contact-details-list">
+<% if contact[:responsible] %>
+  <li><span class="contact-details-list--responsible">Responsible Tenant</span></li>
+<% end %>
   <li>Title: <%= contact[:title] %></li>
   <li>First Name: <%= contact[:first_name] %></li>
   <li>Last Name: <%= contact[:last_name] %></li>
--- a/app/views/common/_contact_details.html.erb
+++ b/app/views/common/_contact_details.html.erb
@@ -1,12 +1,12 @@
 <ul class="list contact-details-list">
 <% if contact[:responsible] %>
-  <li><span class="contact-details-list--responsible">Responsible Tenant</span></li>
+  <li><span class="contact-details-list__responsible">Responsible Tenant</span></li>
 <% end %>
-  <li>Title: <%= contact[:title] %></li>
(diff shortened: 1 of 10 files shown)


app/views/common/_contact_details.html.erb
```diff
@@ -1,4 +1,7 @@
-<ul class="list">
+<ul class="list contact-details-list">
+<% if contact[:responsible] %>
+  <li><span class="contact-details-list--responsible">Responsible Tenant</span></li>
```
zuren: Would consider this an element, not a modifier. Modifiers are only used alongside a base element class (e.g. `button button--green`.

So should be `contact-details-list__responsible`

StevenLeighton21: BEM tricks, ty :)

app/views/tenancies/show.html.erb
```diff
@@ -35,7 +35,8 @@
           <li><%= @tenancy.primary_contact_long_address %></li>
           <li><%= @tenancy.primary_contact_postcode %></li>
         </ul>
-        <% if !@tenancy.contacts[0].nil? %>
+        <% primary_contact = @tenancy.contacts.select { |t| t[:responsible] }.first %>
```
zuren: Not a necessary change, but `list.find { ...predicate... }` is the same as `list.select { ...predicate... }.first`, and a bit faster as it doesn't have to traverse the whole array unnecessarily.

spec/lib/hackney/income/view_tenancy_spec.rb
```diff
@@ -59,6 +59,7 @@
         expect(subject.contacts[0].fetch(:post_code)).to eq('E0 123')
         expect(subject.contacts[0].fetch(:date_of_birth)).to eq('12th March, 1976')
         expect(subject.contacts[0].fetch(:hackney_homes_id)).to eq('1209')
+        expect(subject.contacts[0].fetch(:responsible)).to eq('true')
```
zuren: Think it should cast this to a boolean?

This code in the view is untested, and suspect it might not work correctly, as 'false' is a truthy value:

```rb
@tenancy.contacts.select { |t| t[:responsible] }.first
```

StevenLeighton21: I think the correct thing would be to use boolean values instead of strings in the test, as the API provides a boolean value.

PR #109, opened 2018-10-08
Delete remaining tenancies sync code

cob16:
This is moved to API so can be cleaned up


cob16:
@StevenLeighton21 SMS jobs are still being done on frontend side I believe so that has to get moved first.

PR #110, opened 2018-10-09
Add Search

cob16:
- Add search use case/gateway/model 
- Add new page on /search
- Refactor human_readable_agreement_status into a helper

Hopefully forward compatible with @madetech-jeffpinkham's future change to add page_count to API


app/assets/stylesheets/components/main.scss
```diff
@@ -64,6 +64,28 @@ main {
...
+  }
+
+  .form-label--bold {
+    font-weight: bold;
+  }
+
+  .form-search {
+    margin: 20px;
+
+    .form-search--group {
```
StevenLeighton21: I think that the css for the new elements should go into their own file. We're using BEM so I think `form-search` is the block, `group` is an element so should be nested as `&__group` and `bold` would be a modifier, so `&--bold`. There's existing examples but we can go through it in detail if it's helpful

cob16: like this ```scss

  .form-search {
    margin: 20px;

    .&__label--bold {
        font-weight: bold;
      }
    }

    .&__group {
      margin-top: 4px;
      width: 80%;
      float: left;
    }

    .&__input {
      width: 80%
    }
  }
```

Comment (StevenLeighton21) [reply app/assets/stylesheets/components/main.scss]: I think like this
```css
&__label {
      &--bold {
           font-weight: bold;
      }
}
```
```html
<div class="form-search form-search__label form-search__label--bold">

**Comment by StevenLeighton21** [comment]
@mrosel Can you suggest what, if anything is necessary to ensure we're passing sanitized input to the search API? I'm not hugely sure if there's anything we need to do.

File: app/helpers/agreement_status_helper.rb [hunk 2]
```diff
@@ -0,0 +1,10 @@
+module AgreementStatusHelper
```
StevenLeighton21: Can you update these to match this please?:
image

This seems a good time to do it, even though it's not directly related to this PR

StevenLeighton21: There's a few other areas to consider when doing this, but not directly related to this PR so feel free to raise as issues if need be:
There's a selector that highlights active agreements on the detail page, should highlight `Live` agreements.

app/views/search_tenancies/show.html.erb
```diff
@@ -0,0 +1,53 @@
...
+  <thead>
+  <tr>
+    <th>Ref No.</th>
+    <th>Tenant</th>
+    <th class="numeric">Balance</th>
+    <th>Agreement</th>
+  </tr>
+  </thead>
+  <tbody>
+  <% @results.each do |tenancy| %>
```
StevenLeighton21: Suggest the table row could be a partial, like the payment transaction or arrears action.

config/environments/development.rb
```diff
@@ -1,6 +1,9 @@
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
 
+  # allow console output in docker
+  config.web_console.whitelisted_ips = '10.0.2.2'
```
StevenLeighton21: Should remove this/not commit it.

lib/hackney/income/search_tenancies_gateway.rb
```diff
@@ -0,0 +1,47 @@
...
+        uri.query = URI.encode_www_form(
+          'SearchTerm' => search_term,
+          'Page' => page,
+          'PageSize' => page_size
+        )
+
+        request = Net::HTTP::Get.new(uri)
+        request['X-Api-Key'] = @api_key
+
+        responce = Net::HTTP.start(uri.hostname, uri.port, use_ssl: true) { |http| http.request(request) }
```
StevenLeighton21: should be `response`

lib/hackney/income/search_tenancies_gateway.rb
```diff
@@ -0,0 +1,47 @@
...
+
+        json = JSON.parse(responce.body)
+
+        tenancies = []
+        number_of_pages = json.dig('number_of_pages').to_i
+        json.dig('data', 'tenancies')&.each do |tenancy|
```
StevenLeighton21: I might just parse the data field directly rather than digging (`json = JSON.parse(responce.body)[:data]`) but it's personal preference.

lib/hackney/income/search_tenancies_usecase.rb
```diff
@@ -0,0 +1,17 @@
+module Hackney
+  module Income
+    class SearchTenanciesUsecase
+      def initialize(search_gateway:)
+        @search_gateway = search_gateway
+      end
+
+      def execute(search_term:, page: 0)
```
madetech-jeffpinkham: Suggest setting the default as 10 for pageSize as a parameter?
```def execute(search_term:, page: 0, pageSize:10)```
I foresee users down the line asking for them to be able to change pageSize

Comment (cob16) [reply lib/hackney/income/search_tenancies_usecase.rb]: Its an option in the gateway but I excluded it from the use case/controller etc until we know the performance impact. I am happy to expose it if we agree on a max value.

File: lib/hackney/income/search_tenancies_gateway.rb [hunk 8]
```diff
@@ -0,0 +1,47 @@
...
+module Hackney
+  module Income
+    class SearchTenanciesGateway
+      def initialize(api_host:, api_key:)
+        @api_host = api_host
+        @search_url = URI.parse(File.join(api_host, '/tenancies/search'))
+        @api_key = api_key
+      end
+
+      def search(search_term:, page:, page_size:)
```
mrosel: This method is doing to much - lets split out the request from the results from the building of the response object.

lib/hackney/income/search_tenancies_gateway.rb
```diff
@@ -0,0 +1,47 @@
...
+        uri = @search_url.dup
+
+        uri.query = URI.encode_www_form(
+          'SearchTerm' => search_term,
+          'Page' => page,
+          'PageSize' => page_size
+        )
+
+        request = Net::HTTP::Get.new(uri)
```
mrosel: Could we use a lib here? Faraday or httparty handle json data in a much cleaner way, error handling etc .

spec/features/search_tenancies_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    expect(page).to have_no_content('There are no results')
+    expect(page.results.length).to eq(1)
+  end
+
+  def stub_my_cases_response
+    stub_const('Hackney::Income::IncomeApiUsersGateway', Hackney::Income::StubIncomeApiUsersGateway)
+
+    response_json = File.read(Rails.root.join('spec', 'fixtures', 'my_cases_response.json'))
+    stub_request(:get, "#{ENV['INCOME_COLLECTION_LIST_API_HOST']}/my-cases?number_per_page=20&page_number=1&user_id=1")
```
StevenLeighton21: What are you using as the env value here?  I think the test should still work even if it isn't set, based on the design. So I'd suggest setting it explicitly in your `before` block and deleting in your `after` block.

spec/helpers/agreement_status_helper_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'rails_helper'
+
+describe AgreementStatusHelper do
```
StevenLeighton21: 💯

lib/hackney/income/use_case_factory.rb
```diff
@@ -110,6 +116,13 @@ def events_gateway
         Hackney::Income::SqlEventsGateway.new
       end
 
+      def search_tenancies_gateway
+        Hackney::Income::SearchTenanciesGateway.new(
+          api_host: ENV['INCOME_COLLECTION_API_HOST'],
```
mrosel: `ENV` as an array has a habit of failing silently at runtime. A better method is to use fetch `ENV.fetch('INCOME_COLLECTION_API_HOST')` will fail to boot the app (and tell you what's missing)

spec/lib/hackney/income/search_tenancies_gateway_spec.rb
```diff
@@ -0,0 +1,165 @@
...
+    {
+    search_term: search_term,
+    page: search_term,
+    page_size: page_size
+    }
+  end
+
+  let(:url_params) do
+    {
+    "SearchTerm": params[:search_term],
```
StevenLeighton21: Indentation looks off here - is it just me? Seems odd that rubocop didn't pick it up

app/views/search_tenancies/_form.html.erb
```diff
@@ -0,0 +1,9 @@
+<form method="get" class="form-search" action="<%= search_path %>" >
```
mrosel: I don't particularly care for hand rolled forms. Is there a reason not to use `form_for`?

spec/lib/hackney/income/search_tenancies_gateway_spec.rb
```diff
@@ -0,0 +1,165 @@
...
+        headers: { 'X-Api-Key' => 'skeleton' },
+        query: url_params
+      )
+      expect(request).to have_been_made
+    end
+  end
+
+  private
+
+  def generate_fake_tennancy_results(number)
```
StevenLeighton21: should be `tenancy` :)

spec/lib/hackney/income/search_tenancies_gateway_spec.rb
```diff
@@ -0,0 +1,165 @@
...
+        "prop_ref": Faker::Number.number(8),
+        "tenure": Faker::Lorem.characters(3),
+        "current_balance": {
+          "value": Faker::Number.negative.round(2),
+          "currency_code": 'GBP'
+        },
+        "current_arrears_agreement_status": '100',
+        "primary_contact": {
+          "name": "#{Faker::Name.prefix} #{Faker::Name.name}",
+          "short_address": Faker::Address.street_address.to_s,
```
StevenLeighton21: I think these fields are already strings, so can drop the `to_s`

lib/hackney/income/search_tenancies_usecase.rb
StevenLeighton21: Potentially, this is a minor tweak to allow some maintainability in future, but I'd be happy with not changing it also as it's unused in our current scope.

mrosel:
> @mrosel Can you suggest what, if anything is necessary to ensure we're passing sanitized input to the search API? I'm not hugely sure if there's anything we need to do.

Will need to have a look at this - We aren't searching on discrete fields, so there's not much we can sanitize. However, we should be wary of any logging with regard to personally identifying attributes. I don't see any here ALM

lib/hackney/income/search_tenancies_usecase.rb
madetech-jeffpinkham: Lets go with 50, staging is currently returning 497ms for searchTerm=smith&pageSize=50

config/environments/development.rb
cob16: I put it in the development as I assumed it won't run in prod. I keep having to add it when running the project. Why is this a bad idea?

spec/features/search_tenancies_spec.rb
cob16: - do stubs not get auto torn down?

I think we have faling tests right now if you don't have these vars set already thats somthing i want to get rid of but i think its out of the scope of this pull request.

StevenLeighton21: I meant that if you set the env var in your before block, delete it in your after block. The stub will be, but if your test uses an env I'd suggest setting it explicitly in the test.

config/environments/development.rb
StevenLeighton21: Nothing should be logged to console directly in production code, so it should be redundant. But, if you leave logs in you could potentially leak data this way. Better just not to log to the console at all outside of dev IMO

cob16: Right, that's why I thought to put it in   `environments/development.rb` so that it can't run in production.

PR #111, opened 2018-10-11
Show page count on search page

cob16:
- add pagination

image

### DIFF
--- a/app/controllers/search_tenancies_controller.rb
+++ b/app/controllers/search_tenancies_controller.rb
@@ -2,16 +2,13 @@ class SearchTenanciesController < ApplicationController
   before_action :sanitize_page_params, :no_cache
 
   def show
-    @keyword = search_params[:keyword]
-    @page = search_params[:page]
-    @results = []
-    @number_of_pages = 0
-
-    return unless @keyword
-
-    resp = use_cases.search_tenancies.execute(search_term: @keyword, page: @page)
-    @results = resp[:results]
-    @number_of_pages = resp[:number_of_pages]
+    keyword = search_params[:keyword]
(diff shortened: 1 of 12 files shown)


app/controllers/search_tenancies_controller.rb
```diff
@@ -2,16 +2,13 @@ class SearchTenanciesController < ApplicationController
...
-    @number_of_pages = 0
-
-    return unless @keyword
-
-    resp = use_cases.search_tenancies.execute(search_term: @keyword, page: @page)
-    @results = resp[:results]
-    @number_of_pages = resp[:number_of_pages]
+    keyword = search_params[:keyword]
+    page = search_params[:page]
+    @results = nil
```
mrosel: @results can return multiple types, you may be better served to make it a `Array`, instead of `nil` , would also mean you don't need this more complicated if https://github.com/LBHackney-IT/LBH-IncomeCollection/pull/111/files#diff-9c70ae5ab0b8a8eaf3fa95818fee7b49R3

app/controllers/search_tenancies_controller.rb
```diff
@@ -2,16 +2,13 @@ class SearchTenanciesController < ApplicationController
...
+    if keyword
+      @results = use_cases.search_tenancies.execute(search_term: keyword, page: page)
+    end
+    @search_info = { page: page, keyword: keyword }
```
mrosel: only thing I don't like here is that the controller is doing more than necessary, we are sending two things to the view, `@results` and `@search_info`. I feel like this logic should be handled in the usecase and should only use a `@search_tenancies` (tbd)
... and that search object should hold the search info and results.

PR #112, opened 2018-10-15
Search start at page 1 for application consistancy 

cob16:
The worktray starts on page 1 so the search is now acting the same


spec/controllers/search_tenancies_controller_spec.rb
```diff
@@ -7,46 +7,26 @@
   end
 
   context '#show' do
-    it 'displays an results page when no keyword' do
+    it 'displays an results page when no search term' do
```
StevenLeighton21: `displays a results`, if I'm being _extra_ picky

PR #117, opened 2018-10-23
Rescue omniauth errors, keep failure route

StevenLeighton21:
Testing this issue seems to be a problem - forcing omniauth to redirect to the failure route doesn't seem to work, nor does test_mode actually raise any exceptions. I can use `get :failure` instead of :create, but this feels again like the actual integration is untested.

Rescue on `OmniAuth::Strategies::AzureActiveDirectory::OAuthError` covers what I'm seeing in the logs for access denied errors, but I feel like the test is at this point redundant unless it is properly testing the actual behaviour. Right now, this test is meaningless? 

Suggest just removing it and ad-hoc testing this as the docs for omniauth seem to differ from the actual behaviour. Open to discussion.

EDIT: Now monkeypatched the gem to rescue and redirect to 'auth/failure' as expected. In future, if the gem updates to behave as documented, simply removing the patch should still work.

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -10,10 +10,7 @@ class SessionsController < ApplicationController
   end
 
   def create
-    if auth_hash.extra.raw_info.nil? || !user_has_ic_staff_permissions?
-      flash[:notice] = 'You do not have the required access permission'
-      return redirect_to login_path
-    end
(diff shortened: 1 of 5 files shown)


cob16:
Looks like we are wrapping the lib to catch the exception raised instead of monkey patching so I'm all for it. 👍

PR #118, opened 2018-10-23
Guard against swallowing 500 errors from api

cob16:
- Throw the corresponding IncomeApiError or TenancyApiError when receiving non 200 results for all API gateways
- Create a custom exception so that it's easy to tell what API is throwing errors

### DIFF
--- a/lib/exceptions.rb
+++ b/lib/exceptions.rb
@@ -1,25 +1,25 @@
 module Exceptions
   class IncomeApiError < StandardError
-    attr_reader :responce
-    def initialize(responce)
+    attr_reader :response
+    def initialize(response)
       super
-      @responce = responce
+      @response = response
     end
 
     def to_s
-      "[Income API error: Received #{responce.code} responce] #{super}"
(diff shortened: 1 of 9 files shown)


lib/exceptions.rb
```diff
@@ -0,0 +1,25 @@
+module Exceptions
+  class IncomeApiError < StandardError
+    attr_reader :responce
```
StevenLeighton21: I think it should be `response`

spec/lib/hackney/income/search_tenancies_gateway_spec.rb
```diff
@@ -51,101 +51,117 @@
...
+          query: url_params
+        )
+        expect(request).to have_been_made
+      end
     end
-  end
 
-  context 'when searching for something with no results' do
-    let(:tenancies_results) { [] }
+    context 'when there is some strange json' do
```
StevenLeighton21: Suggest this could be 'when the request returns no results'

PR #121, opened 2018-10-26
Add GPLv3 Licence

cob16:
TLDR on the GPL V3
> You may copy, distribute and modify the software as long as you track changes/dates in source files. Any modifications to or software including (via compiler) GPL-licensed code must also be made available under the GPL along with build & install instructions.


cob16:
Pending further discussion

PR #125, opened 2018-10-30
Ability to pause cases

elena-vi:
screenshot 2018-10-30 at 11 11 31

The ability to pause can be accessed via the `/tenancies/[id]` page:
- Setting the date to a future date pauses the case, meaning that it can only be accessed via the `paused` tab. 
- Setting the date to a past date unpauses the case, meaning that it can be accessed via the worktray

### DIFF
--- a/app/controllers/tenancies_controller.rb
+++ b/app/controllers/tenancies_controller.rb
@@ -15,6 +15,10 @@ class TenanciesController < ApplicationController
     @tenancy = use_cases.view_tenancy.execute(tenancy_ref: params.fetch(:id))
   end
 
+  def pause
+    @tenancy = use_cases.view_tenancy.execute(tenancy_ref: params.fetch(:id))
+  end
+
   private
 
   # FIXME: stop filtering here, improve contact details
--- a/app/controllers/tenancies_controller.rb
+++ b/app/controllers/tenancies_controller.rb
@@ -18,7 +18,13 @@ class TenanciesController < ApplicationController
   end
 
   def update
-    @tenancy = use_cases.view_tenancy.execute(tenancy_ref: params.fetch(:id))
+    use_cases.update_tenancy.execute(
+      tenancy_ref: params.fetch(:id),
+      is_paused_until: params.fetch(:is_paused_until)
+    )
+
+    flash[:notice] = 'Successfully paused'
+    redirect_to tenancy_path(id: params.fetch(:id))
   end
 
   private
--- a/app/controllers/tenancies_controller.rb
+++ b/app/controllers/tenancies_controller.rb
@@ -17,7 +17,7 @@ class TenanciesController < ApplicationController
     @tenancy = use_cases.view_tenancy.execute(tenancy_ref: params.fetch(:id))
   end
 
-  def pause
+  def update
     @tenancy = use_cases.view_tenancy.execute(tenancy_ref: params.fetch(:id))
   end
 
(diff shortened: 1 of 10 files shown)


app/controllers/tenancies_controller.rb
```diff
@@ -17,6 +17,16 @@ def show
     @tenancy = use_cases.view_tenancy.execute(tenancy_ref: params.fetch(:id))
   end
 
+  def update
```
StevenLeighton21: The API function is capable of returning some usefully informative failure responses, suggest you could handle 204 and 500 responses here with different flash messages?

StevenLeighton21:
You should include a description of the changes in the PR, this RFC has a useful outline of what to include https://github.com/madetech/rfcs/blob/master/CodeReviewStandards/rfc-008-review-checklist.md

In the case of PRs with UI elements, a screenshot of the visual changes should also be included.

lib/hackney/income/update_tenancy.rb
```diff
@@ -0,0 +1,16 @@
+require 'date'
+require 'ostruct'
```
StevenLeighton21: Do you need these requires in this file still?

spec/lib/hackney/income/less_dangerous_tenancy_gateway_spec.rb
```diff
@@ -374,6 +374,23 @@
       end
     end
   end
+
+  context 'when updating a tenancy' do
```
StevenLeighton21: Should add a context of `when unsuccessfully updating a tenancy` - you can stub the request to return a [500, :internal_server_error] and handle accordingly.

elena-vi:
[Build basic create pause form with date-picker](https://trello.com/c/9w5GOJdI/23-build-basic-create-pause-form-with-date-picker)

config/routes.rb
```diff
@@ -7,6 +7,7 @@
   get '/search', to: 'search_tenancies#show'
 
   get '/tenancies/:id', to: 'tenancies#show', as: :tenancy
+  patch '/tenancies/:id', to: 'tenancies#update', as: :update
```
mrosel: the `as: :update` is redundant/default as patch always matches to `update` unless overridden

lib/hackney/income/less_dangerous_tenancy_gateway.rb
```diff
@@ -104,6 +104,19 @@ def get_tenancy(tenancy_ref:)
         tenancy_item
       end
 
+      def update_tenancy(tenancy_ref:, is_paused_until:)
+        uri = URI("#{@api_host}/tenancies/#{ERB::Util.url_encode(tenancy_ref)}")
```
cob16: Would like to use this method so that duplcate `/` on api_host are removed

```suggestion
        uri = URI.parse(File.join(@api_host, "/tenancies/#{ERB::Util.url_encode(tenancy_ref)}"))
```

StevenLeighton21: Is it safe to assume a file path and a rest endpoint are interchangeable?

cob16: This seems to be a common usage in rails

StevenLeighton21: 🚢

PR #128, opened 2018-10-31
Remove worker

StevenLeighton21:
Remove the docker image as it can spawn logs that are totally unnecessary.
Document some files that I don't want to remove yet, but will be moved to the IC API in time when events are scheduled on the API.


StevenLeighton21:
Thanks @mrosel, will do!

PR #129, opened 2018-10-31
hotfixing notifications on the client side

elena-vi:
screenshot 2018-10-31 at 11 58 29

Making sure that a successful message is shown when a no content response is received


elena-vi:
Sorry Mark, next time need to :shipit:

PR #135, opened 2018-11-21
Manually send an email message via gov notify

elena-vi:
Introduced the ability to send email to multiple responsible tenants
screenshot 2018-11-21 at 16 00 38


elena-vi:
[Manually send email message via Gov Notify](https://trello.com/c/SIOQSb7s/8-manually-send-email-message-via-gov-notify)

PR #140, opened 2018-11-28
Simple pause part 2

elena-vi:
Adding the ability to set pause comment and pause reason, as well as adding an action into the actions diary once a pause is set 
screenshot 2018-11-28 at 13 39 11


elena-vi:
[Simple pause - Part 2](https://trello.com/c/mHRa4WJ0/28-simple-pause-part-2)

PR #146, opened 2018-12-05
Remove basic auth for production

cob16:
Now that we have OAuth in front of the app the time has come to remove the now redundant basic auth layer for the production environment. 

**As this is an auth change I have added many devs as a heads up approve/reject this PR. Can we all approve this PR to acknowledge this change?**

notes:
 - add 'private_environment' bool config setting
 - set it to false on production app config


cob16:
@StevenLeighton21 keeping it on for staging as we are auto-deploying there (that way auth bugs don't open up the system) and to keep regular users out.

StevenLeighton21:
@cob16 Makes sense! 👍

PR #158, opened 2019-01-29
Ui adaption show pay ref on the case details page

elena-vi:
screenshot 2019-01-29 at 14 06 14

### DIFF
--- a/lib/hackney/income/tenancy_gateway.rb
+++ b/lib/hackney/income/tenancy_gateway.rb
@@ -100,6 +100,7 @@ module Hackney
           t.primary_contact_name = tenancy['tenancy_details']['primary_contact_name']
           t.primary_contact_long_address = tenancy['tenancy_details']['primary_contact_long_address']
           t.primary_contact_postcode = tenancy['tenancy_details']['primary_contact_postcode']
+          t.pay_ref = tenancy['tenancy_details']['pay_ref']
 
           t.arrears_actions = extract_action_diary(events: tenancy['latest_action_diary_events'])
           t.agreements = extract_agreements(agreements: tenancy['latest_arrears_agreements'])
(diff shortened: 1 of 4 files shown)


elena-vi:
[UI Adaption - Show pay ref on the case details page](https://trello.com/c/TyzZ1pEi/103-ui-adaption-show-pay-ref-on-the-case-details-page)

lib/hackney/income/tenancy_gateway.rb
```diff
@@ -100,6 +100,7 @@ def get_tenancy(tenancy_ref:)
           t.primary_contact_name = tenancy['tenancy_details']['primary_contact_name']
           t.primary_contact_long_address = tenancy['tenancy_details']['primary_contact_long_address']
           t.primary_contact_postcode = tenancy['tenancy_details']['primary_contact_postcode']
+          t.pay_ref = tenancy['tenancy_details']['pay_ref']
```
cob16: This will explode for 2 reasons
- the field does not exist
- the field will be called "payment_ref"
- the field is sometimes null anyway

```suggestion
          t.pay_ref = tenancy.dig('tenancy_details', 'payment_ref')
```

cob16: This suggested change means this card can be merged now

spec/lib/hackney/income/tenancy_gateway_spec.rb
```diff
@@ -190,6 +190,7 @@
 
       it 'should return a single tenancy matching the reference given with converted currencies' do
         expect(subject).to be_instance_of(Hackney::Income::Domain::Tenancy)
+        expect(subject.pay_ref).to eq(expected_details.fetch(:pay_ref))
```
cob16: we need a test where this field is nil

PR #162, opened 2019-02-06
Hotfix tenancy graph

elena-vi:
TO DO: 
- [ ] fix capybara to be able to fully test the graph  

screenshot 2019-02-06 at 17 01 09


elena-vi:
[Tenancy Graph causing some pages to 500 in production](https://trello.com/c/nlBrAJR6/120-tenancy-graph-causing-some-pages-to-500-in-production)

spec/features/view_a_case_spec.rb
```diff
@@ -49,6 +49,8 @@ def then_i_should_see_case_account_balance
   def then_i_should_see_balance_graph
     expect(page.body).to have_css('h2', text: 'Balance over time', count: 1)
     expect(page.body).to have_css('canvas#balance_chart', count: 1)
+    # page.execute_script('draw_graph()')
+    # expect(page.body).to have_script('var transactions = []', count: 1)
```
cob16: Has the test been disabled?

alexdesi: do we need this comments?

elena-vi: because we need to fix our test suite to be able to run capybara with js and properly test the page, but because it takes a lot of time to do so and because I think this issue made it to production I'm leaving those comments in there. Hopefully this refactor won't be exiled to the backlog and we could get those tests in soon :)

elena-vi: Screenshot added to description ⬆️

PR #164, opened 2019-02-07
removed lots of loops and got ordering time to a second

elena-vi:
from 9 seconds loading to 1 💪

### DIFF
--- a/lib/hackney/income/transactions_balance_calculator.rb
+++ b/lib/hackney/income/transactions_balance_calculator.rb
@@ -2,22 +2,20 @@ module Hackney
   module Income
     class TransactionsBalanceCalculator
       def organise_with_final_balances_by_week(current_balance:, transactions:)
-        weeks = get_weeks(transactions)
+        weeks_and_transactions = get_weeks(transactions)
+
         transactions_summary = []
 
-        weeks.each do |week|
-          transactions_in_week = get_transactions_by_week(transactions, week)
-          incoming_sum = get_incoming_sum(transactions_in_week)
-          outgoing_sum = get_outgoing_sum(transactions_in_week)
-
-          next unless transactions_in_week.any?
-          transactions_summary << {
-            week: week,
-            incoming: incoming_sum,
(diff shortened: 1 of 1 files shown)


lib/hackney/income/transactions_balance_calculator.rb
```diff
@@ -2,19 +2,17 @@ module Hackney
...
+
           transactions_summary << {
             week: week,
-            incoming: incoming_sum,
-            outgoing: outgoing_sum,
+            incoming: get_incoming_sum(transactions_in_week),
+            outgoing: get_outgoing_sum(transactions_in_week),
             summarised_transactions: transactions_in_week,
             final_balance: weekly_balance(current_balance, transactions_summary)
           }
```
alexdesi: That's a nice improvement!
we can go further:
- remove the `transaction_summary array`
- remove the each and use a `map`
- remove the guard clause
Something like this:
```
weeks_with_transactions = weeks_and_transactions.select {|_week, transactions_in_week| transactions_in_week.any?| }

weeks_with_transactions.map do |week, transactions_in_week|
    {
      week: week,
      incoming: incoming_sum,
      outgoing: outgoing_sum,
      incoming: get_incoming_sum(transactions_in_week),
      outgoing: get_outgoing_sum(transactions_in_week),
      summarised_transactions: transactions_in_week,
      final_balance: weekly_balance(current_balance, transactions_summary)
	}
end
```

alexdesi: This is a common pattern, ofter we can replace, each with conditions inside, with `select` + `map` (this is a bit of more functional way,  but if you don't like feel free to ignore this 😉  )

elena-vi: Actually this is not possible because `weekly_balance` looks for the last transaction week balance to calculate the next weeks transaction balance

alexdesi: thanks for trying 👍

PR #165, opened 2019-02-07
correct decimal places for sms and testing

elena-vi:
screenshot 2019-02-07 at 18 15 44


elena-vi:
[Send SMS missing a decimal place e.g SMS should be £2024.40 but is displaying as £2024.4](https://trello.com/c/AscraIdz/119-send-sms-missing-a-decimal-place-eg-sms-should-be-202440-but-is-displaying-as-20244)

elena-vi:
to do: 

- [ ] change up the way tenancy balance is returned everywhere

PR #166, opened 2019-02-26
Ui send a letter

elena-vi:


elena-vi:
[Step 1 - UI send a Letter](https://trello.com/c/GvGYzdkE/153-step-1-ui-send-a-letter)

app/controllers/letters_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+    @letter_templates = use_cases.list_letter_templates.execute
+  end
+
+  def preview
+    @preview = use_cases.get_letter_preview.execute(
+      template_id: params[:template_id],
+      pay_ref: params[:pay_ref],
+      user_id: session[:current_user].fetch('id')
+    )
+  rescue Exceptions::IncomeApiError::NotFoundError
```
cob16: I am worried about the coupling of the 3 layers here. This exception is thrown all the way back in the gateway. I Suggest we handle this either by the gateway returning something else (like an empty array) or by catching it in the use case and returning a hash of some sort.

app/views/letters/preview.erb
```diff
@@ -0,0 +1 @@
+<%= @preview[:preview].html_safe %>
```
cob16: It seems odd that the backend is doing the entire HTML rendering for the frontend here but I don't see any other way here.

lib/hackney/income/letters_gateway.rb
```diff
@@ -0,0 +1,45 @@
...
+          payment_ref: payment_ref,
+          template_id: template_id,
+          user_id: user_id
+        }.to_json
+
+        req = Net::HTTP::Post.new(uri)
+        req['X-Api-Key'] = @api_key
+        req['Content-Type'] = 'application/json'
+
+        res = Net::HTTP.start(uri.host, uri.port, use_ssl: true) { |http| http.request(req, body_data) }
```
cob16: The usual "Would be nice if we used HTTP party here"

elena-vi: yea.. @mrosel thinks so too. I think we're gonna be adding this to the list of things to refactor soonish

app/views/letters/preview.erb
elena-vi: 🤷‍♀️

mrosel: I struggled with this at first. But the api is returning a json data structure with a `Preview`. The type of the `Preview` is `html` / string, so it's still up to the front end to know how to handle preview types. This could easily be changed to a binary (pdf) or image later but that likely won't happen.

app/controllers/letters_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+      template_id: params.fetch(:template_id),
+      pay_ref: params.fetch(:pay_ref),
+      user_id: session[:current_user].fetch('id')
+    )
+    if @preview[:status_code] == 404
+      flash[:notice] = 'Payment reference not found'
+      redirect_to letters_new_path
+    end
+  rescue ActionController::ParameterMissing
+    flash[:notice] = 'Error fetching preview: Please enter a payment reference'
```
mrosel: MUST not rescue without followed by a log/sentry/raise

lib/hackney/income/get_letter_preview.rb
```diff
@@ -0,0 +1,19 @@
...
+      end
+
+      def execute(pay_ref:, template_id:, user_id:)
+        @letters_gateway.send_letter(
+          payment_ref: pay_ref,
+          template_id: template_id,
+          user_id: user_id
+        )
+      rescue Exceptions::IncomeApiError::NotFoundError
+        { status_code: 404 }
```
mrosel: the same pattern as before - MUST not rescue without being followed by a log/sentry/raise

lib/hackney/income/letters_gateway.rb
```diff
@@ -0,0 +1,45 @@
+module Hackney
+  module Income
+    class LettersGateway
+      def initialize(api_host:, api_key:)
+        @api_host = api_host
+        @api_key = api_key
+      end
+
+      def send_letter(payment_ref:, template_id:, user_id:)
+        uri = URI("#{@api_host}v1/pdf/send_letter")
```
mrosel: suggest we move `v1/pdf/send_letter` to a constant of `LettersGateway`. This makes refactoring and version changes easier

lib/hackney/income/use_case_factory.rb
```diff
@@ -129,6 +141,13 @@ def notifications_gateway
         )
       end
 
+      def letters_gateway
+        Hackney::Income::LettersGateway.new(
+          api_host: INCOME_API_HOST,
+          api_key: ENV['INCOME_COLLECTION_API_KEY']
```
mrosel: suggest using `ENV.fetch('INCOME_COLLECTION_API_KEY')`

spec/controllers/letters_controller_spec.rb
```diff
@@ -0,0 +1,63 @@
+require 'rails_helper'
+
+describe LettersController do
+  before do
+    stub_authentication
+  end
+
+  let(:template_id) { '123456789' }
```
mrosel: could we be using faker here too?

template_id, template_name, and payment_ref?

spec/controllers/letters_controller_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+      expect(assigns(:letter_templates)).to all(be_valid)
+    end
+  end
+
+  context '#preview' do
+    it 'shows preview' do
+      expect_any_instance_of(Hackney::Income::LettersGateway).to receive(:send_letter).with(
+        payment_ref: payment_ref,
+        template_id: template_id,
+        user_id: 123
```
mrosel: suggest you replace these `user_id` with a variable. At some point, we will want to validate the user is the same from `stub_authentication`

elena-vi: If it's not `123` the test fails :) https://github.com/LBHackney-IT/LBH-IncomeCollection/blob/a436989bfc9e21dc19130c969e76d0edf32fbcf0/spec/helper.rb

spec/controllers/letters_controller_spec.rb
elena-vi: yea, my bad

spec/controllers/letters_controller_spec.rb
cob16: perhaps we add this as a constant?

spec/lib/hackney/income/list_letter_templates_spec.rb
```diff
@@ -0,0 +1,52 @@
+require 'rails_helper'
+
+describe Hackney::Income::ListLetterTemplates do
+  let(:letters_gateway) { instance_double(Hackney::Income::LettersGateway) }
+
+  subject { described_class.new(letters_gateway: letters_gateway) }
+
+  it 'should use the notification gateway' do
```
alexdesi: Are we using the `should` style or as suggested in better spec we use present tense? 
`it 'uses the notification gatway`?
cc @mrosel @cob16

PR #167, opened 2019-03-11
Preview validation

elena-vi:


elena-vi:
[Step 2 - UI preview Letter](https://trello.com/c/jjZPk7OR/154-step-2-ui-preview-letter)

spec/features/view_letter_preview_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+    stub_get_templates_response
+    stub_post_send_letter_response
+  end
+
+  scenario do
+    given_i_am_logged_in
+    when_visit_new_letter_page
+    then_i_see_a_letter_form
+    then_i_fill_in_the_form
+    then_i_letter_preview_page
```
alexdesi: this reads weird,
maybe you miss something befor `...letter_preview_page`
possibly `then_i_see_the_letter_preview_page`

spec/features/view_letter_preview_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+    stub_my_cases_response
+    stub_get_templates_response
+    stub_post_send_letter_response
+  end
+
+  scenario do
+    given_i_am_logged_in
+    when_visit_new_letter_page
+    then_i_see_a_letter_form
+    then_i_fill_in_the_form
```
alexdesi: you might want to be more explicit here, `then_i_fill_in_the_form_and_confirm` or
`then_i_fill_in_the_form_and_click_preview`

spec/features/view_letter_preview_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+
+  scenario do
+    given_i_am_logged_in
+    when_visit_new_letter_page
+    then_i_see_a_letter_form
+    then_i_fill_in_the_form_and_submit
+    then_i_see_the_letter_preview_with_errors
+  end
+
+  def given_i_am_logged_in
```
mrosel: IMO the ad should be tested in isolation and this doesn't feel relevant to what we're testing. Can we remove this and start our test past this point?
In other words, stub the session and start at `when_visit_new_letter_page`

app/assets/stylesheets/components/letter_preview.scss
```diff
@@ -0,0 +1,47 @@
+.letter_preview {
+  border: 2px solid $hackney-green;
+  padding: 20px;
+  max-height: 50vh;
+  overflow: scroll;
+
+  .right{
```
cob16: currently, the frontend is using the BEM naming scheme. We should stay consistent to that.

So in this example we do
```suggestion
  .&--right{
```
that creates `.letter_preview--right`

app/assets/stylesheets/components/letter_preview.scss
```diff
@@ -0,0 +1,47 @@
...
+    float: right;
+    text-align: right;
+  }
+
+  .top_space {
+    height: 69.12pt;
+  }
+
+  .hackney-logo {
```
cob16: this can be an element in bem i.e `letter-preview__logo`

app/assets/stylesheets/components/letter_preview.scss
```diff
@@ -0,0 +1,47 @@
...
+
+  p {
+    display: block;
+    margin-block-start: 14.4pt;
+    margin-block-end: 14.4pt;
+    margin-inline-start: 0px;
+    margin-inline-end: 0px;
+  }
+
+  .green_box {
```
cob16: - block is `letter-preview`
- element is the `box`
- modifier is `green`

so with bem that makes it`.letter-preview__box--green`

```suggestion
  .&__box {
    .&--green {
```

elena-vi:
@cob16 I agree with the css comments, but we need to keep in mind that the preview html is coming from the back end where it wouldn't make sense to have things have classes on elements like `.letter-preview__box--green`

spec/features/view_letter_preview_spec.rb
elena-vi: https://trello.com/c/9omLIkmG/169-refactor-feature-test-to-start-where-they-really-start

Backlogged

cob16:
> @cob16 I agree with the css comments, but we need to keep in mind that the preview html is coming from the back end where it wouldn't make sense to have things have classes on elements like `.letter-preview__box--green`

How so?

PR #168, opened 2019-03-29
Step 3 send to notify

elena-vi:


elena-vi:
[Step 3 - Send to Gov Notify](https://trello.com/c/q7vnCCJJ/156-step-3-send-to-gov-notify)

app/controllers/letters_controller.rb
```diff
@@ -12,4 +12,15 @@ def preview
...
     redirect_to letters_new_path if @preview[:status_code] == 404
   end
+
+  def send_letter
+    response = use_cases.send_letter.execute(
+      uuid: params.require(:uuid),
+      user_id: session[:current_user].fetch('id')
+    )
+
+    flash[:notice] = 'Successfully sent' if response.code.to_i == 204
```
cob16: Clean Architecture suggestion incoming 🔥 💣 

Would recommend passing a success=true to the controller from the use case. the actual status code parsing likely should live in the gateway. IMO This line is adding direct coupling from gateway to controller

config/routes.rb
```diff
@@ -8,6 +8,7 @@
 
   get '/letters/new', to: 'letters#new'
   post '/letters/preview', to: 'letters#preview', as: :letter_preview
+  post '/letters/send', to: 'letters#send_letter', as: :send_letter
```
cob16: Not very important but the `/send` part doesn't need to be there. A POST to letters creates a letter (that gets sent).

app/controllers/letters_controller.rb
elena-vi: this is done in several places so I'm adding a tech debt card to refactor where this is being done

PR #174, opened 2019-04-25
Mr ajax send with updates

mrosel:
Extends @elena-vi work with ajax sending.


app/assets/javascripts/preview.js
```diff
@@ -38,3 +50,30 @@ function showLoader() {
...
+    return true;
+  } else {
+    console.log('all send buttons hidden')
+    return false;
+  }
+}
+
+async function submit_send_all_letters(e){
+  var retVal = confirm('Are you sure you want to send all the letters listed here?')
+  if (retVal != true){ return false; }
```
alexdesi: are we using ';' or we are not?

app/controllers/letters_controller.rb
```diff
@@ -33,14 +33,18 @@ def ajax_preview
   end
 
   def send_letter
-    response = use_cases.send_letter.execute(
-      uuid: params.require(:uuid),
+    @letter_uuid = params.require(:uuid)
```
alexdesi: do we need an instance variable here `@letter_uuid`? or just a variable? `letter_uuid`

mrosel: :+1: good catch - this was a hold over from past functionality and should be removed.

app/controllers/letters_controller.rb
```diff
@@ -33,14 +33,18 @@ def ajax_preview
...
+    sent_letter = use_cases.send_letter.execute(
+      uuid: @letter_uuid,
       user_id: session[:current_user].fetch('id')
     )
 
-    flash[:notice] = 'Successfully sent' if response.code.to_i == 204
-
-    redirect_to letters_new_path
+    respond_to do |format|
+      if sent_letter.code.to_i == 204
```
alexdesi: it looks the condition should be inside the block of `format.html`

spec/controllers/letters_controller_spec.rb
```diff
@@ -100,18 +99,35 @@
...
         uuid: uuid,
         user_id: user_id
       ).and_return(Net::HTTPOK.new(1.1, 204, nil))
+    end
 
-      post :send_letter, params: {
-        uuid: uuid,
-        user_id: user_id
-      }
+    context 'html' do
```
alexdesi: `when format is html` is more clear

spec/controllers/letters_controller_spec.rb
```diff
@@ -100,18 +99,35 @@
...
+          uuid: uuid,
+          user_id: user_id
+        }
+        expect(response).to redirect_to(letters_new_path)
+        expect(flash[:notice]).to eq('Successfully sent')
+      end
+    end
 
-      expect(flash[:notice]).to eq('Successfully sent')
+    context 'js' do
```
alexdesi: `when format is js`

app/controllers/letters_controller.rb
mrosel: This is right the `sent_letter` is from the response of the api. 
It's nested under `respond_to` because you usually want a separate response block it the status is failed.

PR #178, opened 2019-04-29
Hf counter spinner

mrosel:
A simple as the api returns a 404 (as expected) when a ref cannot be found. This update hides the spinner whether all success or error or combination of the two.


alexdesi:
@mrosel does it fix the case where the non-exitent payref is in the first position?

PR #180, opened 2019-04-29
flip errors fields and re-align the second

mrosel:
flip errors fields and re-align the n+1 errors


mrosel:
> I am just wondering why there was a ` <td> </td>`
because we added `rowspan` but didn't have a proper test case with multiple errors. :face_with_head_bandage:

PR #182, opened 2019-05-01
Mr async for bugfix

mrosel:
addresses the cause of https://trello.com/c/4rZqoaf3/217-internal-server-error-bug
by queueing the requests so we don't overwhelm the db.

### DIFF
--- a/app/assets/javascripts/preview.js
+++ b/app/assets/javascripts/preview.js
@@ -25,6 +25,13 @@ function handlePreview(){
   }
 }
 
+function handleError(pay_ref, textStatus){
+  handlePreview()
+  increment_fail_counter()
+  $("#errors_table").append("<tr><td>"+pay_ref+"</td><td colspan='2'>"+textStatus+"</td></tr>");
+}
+
+
 function ajax_preview(pay_ref, template_id, max){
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/preview.js
```diff
@@ -18,24 +25,45 @@ function handlePreview(){
...
-      handlePreview()
-      increment_fail_counter()
-      $("#errors_table").append("<tr><td>"+pay_ref+"</td><td colspan='2'>"+response+"</td></tr>");
+    error: function(jqxhr, textStatus, errorThrown){
+      if(errorThrown.status == 500) {
+        this.tryCount++;
+        // retry
+        if (this.tryCount <= this.retryLimit) {
+          Rails.ajax(this);
+          return;
```
alexdesi: not sure if we need this return

app/assets/javascripts/preview.js
```diff
@@ -18,24 +25,45 @@ function handlePreview(){
...
+        } else {
+          console.log(textStatus)
+          handleError(pay_ref, textStatus)
+        }
+        return;
```
alexdesi: I am not even sure if we need this :)

PR #183, opened 2019-05-01
Handling first not found pay ref

elena-vi:


elena-vi:
[Payment ref not found](https://trello.com/c/pj3LOgV6/218-payment-ref-not-found)

app/controllers/letters_controller.rb
```diff
@@ -7,12 +7,19 @@ def new
...
+    @not_founds = []
 
-    @preview = use_cases.get_letter_preview.execute(
-      template_id: params.require(:template_id),
-      pay_ref: @payment_refs.delete_at(0),
-      user_id: session[:current_user].fetch('id')
-    )
+    @preview = generate_letter_preview(payment_ref)
+
+    until @preview[:preview].present? || @payment_refs.empty?
```
alexdesi: hey Elena, your code looks correct,
but it would be cleaner with just a `@payment_refs.each ...` instead of a 
`until` loop.

app/controllers/letters_controller.rb
```diff
@@ -7,12 +7,19 @@ def new
...
+      @not_founds << {
+        payment_ref: payment_ref,
+        error: 'Not Found'
+      }
+      payment_ref = @payment_refs.delete_at(0)
+      @preview = generate_letter_preview(payment_ref)
+    end
 
     flash[:notice] = 'Payment reference not found' if @preview[:status_code] == 404
```
mrosel: is this still relevant? Might need to have a better handling here

app/views/letters/preview.erb
```diff
@@ -1,64 +1,82 @@
...
+          </div>
+        </div>
+      </div>
 
-</div>
+      <div class="grid-row">
+        <div class="column-full">
+          <h2>Unable to send</h2>
+          <h3>
+            Failed count: <span id="failed_count"><%= @preview[:errors].any? ? 1 : 0 %></span>
```
alexdesi: Can `Failed count` be just `1` or `0`?

app/views/letters/preview.erb
```diff
@@ -1,64 +1,82 @@
...
+              <%= render partial: 'errors_table', preview: @preview %>
+            </tbody>
+          </table>
+        </div>
+      </div>
+      <div class="grid-row">
+        <div class="column-full">
+          <h2>Ready to send</h2>
+          <h3>
+            Successful count: <span id="successful_count"><%= @preview[:errors].empty? ? 1 : 0 %></span>
```
alexdesi: same as above

app/views/letters/preview.erb
elena-vi: the javascript increments this as it does down the list

app/views/letters/preview.erb
elena-vi: the javascript increments this as it does down the list

app/controllers/letters_controller.rb
```diff
@@ -7,12 +7,19 @@ def new
 
   def preview
```
mrosel: Suggest: simplifying this

```
    @payment_refs = payment_refs
    @payment_refs.each do |payment_ref|
      @preview = generate_letter_preview(payment_ref)
      next unless @preview[:preview].present?
    end

    if !@preview[:preview].present? || @payment_refs.empty? || @preview[:status_code] == 404
      flash[:notice] = 'None of the payment references were not found'
      redirect_to letters_new_path
    end
```

app/views/letters/preview.erb
```diff
@@ -1,64 +1,82 @@
...
+          <table class="error">
+            <thead>
+              <tr>
+                <th>Payment Ref</th>
+                <th>Error Message</th>
+                <th>Error Field</th>
+              </tr>
+            </thead>
+            <tbody id="errors_table">
+              <% if @not_founds.any? %>
```
mrosel: This isn't needed. See above suggestions

PR #184, opened 2019-05-02
adding buttons, yay!

mrosel:


app/views/tenancies/index.html.erb
```diff
@@ -32,7 +32,11 @@
 
   <% if @user_assigned_tenancies.none? %>
     <h3 class="tenancy_list tenancy_list__no_tenancies">You currently have no assigned cases</h3>
-    <%= link_to 'Send letters', letters_new_path, class:'button' %>
+    <div class='other_actions'>
+      <p> Did you want to perform other actions?</p>
```
alexdesi: nitpick: space after `<p>`

alexdesi: Could you verify the copy? 
Why we are using the past tense? `Did you want...`
do we need that message at all?

mrosel: Removed. Though it felt better with it.

elena-vi:
Actually, could you centre them...? This lonely button looks and the two buttons there will just look awkward 
Screenshot 2019-05-02 at 15 06 18

PR #185, opened 2019-05-02
Fixes bug and order by correct time desc

mrosel:


lib/hackney/income/documents_gateway.rb
```diff
@@ -34,10 +34,10 @@ def get_all
         end
 
         JSON.parse(res.body).map(&:deep_symbolize_keys).each do |doc|
-          doc[:created_at] = Time.parse(doc[:created_at])
-          doc[:updated_at] = Time.parse(doc[:updated_at])
+          doc[:created_at] = doc[:created_at].to_time
+          doc[:updated_at] = doc[:updated_at].to_time
           doc[:metadata] = JSON.parse(doc[:metadata] || '{}').deep_symbolize_keys
-        end
+        end.reverse
```
alexdesi: Mark I am aware I are run out of time.
I am keen to accept this but these are some thoughts:
- maybt this transformations should be done in the `income-api`,
- it's not clear what we are sorting for
- it's not clear what `reverse` is reversing?
- reverse is working in the whole set (heavy computation)
but as I said let's be practical, if you can't handle those atm, we should consider some points as tech debt

mrosel: Yeah, we'll move this to the api. The api returns the data via ascending order which isn't very useful.
For future ref: https://ruby-doc.org/core-2.2.0/Array.html#method-i-reverse

PR #193, opened 2019-07-24
Sort letters in drop down list 

booyaa:
Related Trello card: https://trello.com/c/VZLdFDBN/235-re-ordering-and-rename-of-letters-in-template

### DIFF
--- a/app/views/letters/new.erb
+++ b/app/views/letters/new.erb
@@ -11,7 +11,9 @@
       <div class="form-group">
         <label class="form-label" for="sms-template-select"><strong>Select template</strong></label>
         <span class="form-hint">Select a letter template to send from the dropdown list below</span>
-        <%= select_tag(:template_id, options_for_select(@letter_templates.map { |template| [template.name, template.id]}.sort { |a, b| a[1] <=> b[1]}), id: 'sms-template-select', class:'form-control') %>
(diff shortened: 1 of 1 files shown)


app/views/letters/new.erb
```diff
@@ -11,7 +11,7 @@
       <div class="form-group">
         <label class="form-label" for="sms-template-select"><strong>Select template</strong></label>
         <span class="form-hint">Select a letter template to send from the dropdown list below</span>
-        <%= select_tag(:template_id, options_for_select(@letter_templates.map { |template| [template.name, template.id]}), id: 'sms-template-select', class:'form-control') %>
+        <%= select_tag(:template_id, options_for_select(@letter_templates.map { |template| [template.name, template.id]}.sort { |a, b| a[1] <=> b[1]}), id: 'sms-template-select', class:'form-control') %>
```
alexdesi: It looks good, but (if you got time) you might want to test this:
```
@letter_templates
  .sort_by { |k, _| template.name }
  .map { |template| [template.name, template.id]}
```
 (and if you got more time it should be nice to refactor that into `def ordered_list` inside the `LetterTemplate` model)

cob16:
yeah suggest this be done in the use case so we can test it

PR #201, opened 2019-09-17
call income api to get case priority and show who cases are assigned to

elena-vi:
https://hackney.atlassian.net/browse/MA-104

Screenshot 2019-09-17 at 17 43 52


Soraya300:
Looks good to me

lib/hackney/income/domain/tenancy.rb
```diff
@@ -19,6 +19,14 @@ def display_start_date
           return '' if start_date.nil?
           Time.parse(start_date).to_formatted_s(:long_ordinal)
         end
+
+        def display_assigned_to
+          if case_priority
```
georgeschena: I like `unless` more than `if` in this case

lib/hackney/income/domain/tenancy.rb
```diff
@@ -19,6 +19,14 @@ def display_start_date
           return '' if start_date.nil?
           Time.parse(start_date).to_formatted_s(:long_ordinal)
         end
+
+        def display_assigned_to
```
kjdchapman: I don't understand why some of these fields prefixed with `display`?

lib/hackney/income/use_case_factory.rb
```diff
@@ -166,6 +167,13 @@ def letters_gateway
         )
       end
 
+      def case_priority_gateway
```
maysakanoni: This is the same method as `income_api_tenancy_gateway`. We're also using the same api host and keys.

elena-vi: good catch, that was duplication

spec/examples/single_case_priority_response.json
```diff
@@ -0,0 +1,41 @@
+{
```
kjdchapman: Nice, I like this in a separate file.

spec/features/view_a_case_spec.rb
```diff
@@ -38,6 +38,7 @@ def then_i_should_see_case_meta_data
     expect(page.body).to have_css('li', text: 'Reference number: 1234567/01', count: 1)
     expect(page.body).to have_css('li', text: 'Payment reference: 1010101010', count: 1)
     expect(page.body).to have_css('li', text: 'Start date: August 30th, 2014', count: 1)
+    expect(page.body).to have_css('li', text: 'Assigned to: Billy Bob (Credit Controller)', count: 1)
```
kjdchapman: I understand this is using the same pattern as the other assertions here, but I'd be happy to just check the page's `content` for this.

I think checking the css specifically makes this test more brittle, and doesn't gain us more confidence the feature's working.

spec/features/view_a_case_spec.rb
```diff
@@ -154,6 +155,9 @@ def stub_my_cases_response
   end
 
   def stub_view_case_response
+    response_json = JSON.parse(File.read(Rails.root.join('spec', 'examples', 'single_case_priority_response.json')))
```
kjdchapman: I found this nesting level hard to read - it might be better across multiple lines?

spec/features/view_a_case_spec.rb
```diff
@@ -154,6 +155,9 @@ def stub_my_cases_response
...
+    allow_any_instance_of(Hackney::Income::TenancyGateway).to receive(:get_case_priority).and_return(response_json.deep_symbolize_keys)
```
kjdchapman: Interested to talk about what kind of spec these are in "features" - this having to mock out a constant feels kind of painful?

spec/lib/hackney/income/tenancy_gateway_spec.rb
```diff
@@ -407,6 +407,92 @@
...
+        role: 'base_user'
+      }
+    }
+    end
+
+    subject do
+      tenancy_gateway.get_case_priority(tenancy_ref: case_priority[:tenancy_ref])
+    end
+
+    context 'when there is an assiged case priority' do
```
kjdchapman: nit: typo
```suggestion
    context 'when there is an assigned case priority' do
```

spec/lib/hackney/income/tenancy_gateway_spec.rb
```diff
@@ -407,6 +407,92 @@
...
+      before do
+        stub_request(:get, "https://example.com:443/api/v1/tenancies/#{ERB::Util.url_encode(case_priority[:tenancy_ref])}")
+          .to_return(status: [200], body: case_priority.to_json)
+      end
+
+      it 'get\'s a case priority' do
```
kjdchapman: nit: typo
```suggestion
      it 'gets a case priority' do
```

spec/features/view_a_case_spec.rb
elena-vi: I'm not happy with this line to begin with... 

I'd like to refactor this all of these tests

spec/features/view_a_case_spec.rb
elena-vi: ditto to the comment before..

spec/lib/hackney/income/tenancy_gateway_spec.rb
elena-vi: ❤️

spec/lib/hackney/income/tenancy_gateway_spec.rb
elena-vi: ❤️

PR #204, opened 2019-09-19
Return to last page used after viewing a case

georgeschena:
WHAT: Add a back method to the `Return back to your worktray` link.

WHY: To allow a user to go back to the page they came from and not back to page 1.

### DIFF
--- a/app/views/tenancies/show.html.erb
+++ b/app/views/tenancies/show.html.erb
@@ -2,7 +2,7 @@
 
 <div class="grid-row">
   <div class="column-full">
-    <%= link_to('Return back to your worktray', root_path, class: 'link--back') %>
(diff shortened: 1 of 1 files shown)


elena-vi:
We all can agree that tdd is the one true religion, but I'd like to defend the lack of tests here by saying:
> I find this sort of functionality is unnecessary hard, especially for this little function.
I'm happy to have my mind changed here, if people disagree

maysakanoni:
Looks good! Not sure how you would tdd this 🤔

georgeschena:
So after speaking to others, I feel as this is such a small feature and testing JS is pretty painful we should avoid testing this as it will take a bit of time which can be spent delivering value elsewhere. I have tested this within the browser and it 100% works. In addition I also tested it with JS turned off and it also worked without affecting/damaging the users experience.

PR #205, opened 2019-09-23
Add pagination 

georgeschena:
WHAT: Added pagination 

WHY: To allow easier navigation of tenancies

Screenshot 2019-09-23 at 11 05 12

### DIFF
--- a/app/controllers/search_tenancies_controller.rb
+++ b/app/controllers/search_tenancies_controller.rb
@@ -9,6 +9,7 @@ class SearchTenanciesController < ApplicationController
     @post_code = search_params[:post_code]
     @tenancy_ref = search_params[:tenancy_ref]
 
+
     @results = use_cases.search_tenancies.execute(
       page: page,
       first_name: @first_name,
@@ -17,6 +18,8 @@ class SearchTenanciesController < ApplicationController
       post_code: @post_code,
       tenancy_ref: @tenancy_ref
     )
+    @number_of_pages = @results.dig(:number_of_pages)
+    @tenancies = Kaminari.paginate_array(@results.dig(:tenancies)).page(params[:page]).per(20)
   end
 
   private
--- a/app/controllers/search_tenancies_controller.rb
+++ b/app/controllers/search_tenancies_controller.rb
@@ -9,7 +9,6 @@ class SearchTenanciesController < ApplicationController
     @post_code = search_params[:post_code]
     @tenancy_ref = search_params[:tenancy_ref]
 
-
     @results = use_cases.search_tenancies.execute(
       page: page,
       first_name: @first_name,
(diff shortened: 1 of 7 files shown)


app/controllers/tenancies_controller.rb
```diff
@@ -14,6 +14,8 @@ def index
     @number_of_pages = response.number_of_pages
     @user_assigned_tenancies = valid_tenancies(response.tenancies)
     @showing_paused_tenancies = response.paused
+
+    @tenancies = Kaminari.paginate_array(@user_assigned_tenancies).page(params[:page]).per(10)
```
elena-vi: are you sure this is definitely 10 per page?
I think this might be defined somewhere else

app/controllers/search_tenancies_controller.rb
```diff
@@ -17,6 +17,8 @@ def show
       post_code: @post_code,
       tenancy_ref: @tenancy_ref
     )
+    @number_of_pages = @results.dig(:number_of_pages)
+    @tenancies = Kaminari.paginate_array(@results.dig(:tenancies)).page(params[:page]).per(20)
```
elena-vi: are you sure this is definitely 20 per page?
I think this might be defined somewhere else

kjdchapman: Is this worth splitting up into multiple lines? I found it hard to parse.

```suggestion
    tenancies = @results.dig(:tenancies)
    page = params[:page]
    @tenancies = Kaminari.paginate_array(tenancies).page().per(20)
```

app/views/common/_pagination.html.erb
```diff
@@ -1,18 +1,7 @@
...
-    <div class="column-one-third column-align-center">
-      <span class="pagination__description">Page <%= page_number %> of <%= number_of_pages %></span>
-    </div>
-    <div class="column-one-third column-align-right">
-      <% next_link_href = page_number == number_of_pages ? 'javascript:;' : next_page_path %>
-      <% next_link_class = page_number == number_of_pages ? 'link--disabled link--forward' : 'link--forward' %>
-
-      <%= link_to('Next', next_link_href, class: next_link_class) %>
+    <div class="column-align-center">
+      <%= paginate @tenancies, :total_pages => @number_of_pages %>
```
kjdchapman: image
:ok_hand:

kjdchapman: Oh, could use terser hash syntax here:
```suggestion
      <%= paginate @tenancies, total_pages: @number_of_pages %>
```

app/controllers/search_tenancies_controller.rb
georgeschena: Yes you're correct, I have removed it from the variable and let me be handled elsewhere. Good spot.

app/controllers/tenancies_controller.rb
georgeschena: Yes you're correct, I have removed it from the variable and let me be handled elsewhere. Good spot.

app/controllers/search_tenancies_controller.rb
georgeschena: I feel it reads fairly well now, and I don't want to clutter the controller with more variables.

PR #206, opened 2019-09-23
Hide and show graph MA-110

Miles-Alford:
WHAT: Added hide and show graph on case details page.

WHY: So that the user can tailor the case detail page to suit their preference.

Screenshot 2019-09-23 at 15 07 21

### DIFF
--- a/app/views/tenancies/show.html.erb
+++ b/app/views/tenancies/show.html.erb
@@ -85,8 +85,6 @@
   </div>
 </div>
 
-
-
 <div class="grid-row">
   <div class="column-full">
     <h2>Arrears Agreements</h2>
--- a/app/views/tenancies/show.html.erb
+++ b/app/views/tenancies/show.html.erb
@@ -75,7 +75,7 @@
 
 <div class="grid-row">
   <div class="column-full">
-    <h2>Balance over time</h2>
+    <h2>Balance over time - <a><u id= "hide_graph">click to hide graph</u></a></h2>
     <canvas id="balance_chart" height="300"></canvas>
   </div>
 </div>
@@ -188,6 +188,7 @@
 
 <script type="text/javascript">
   $(document).on('turbolinks:load', draw_graph);
+  $("#hide_graph").on('click', hide_graph);
 
   function draw_graph() {
     {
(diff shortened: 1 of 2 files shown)


georgeschena:
Could you add a screenshot of the UI please!

spec/features/view_a_case_spec.rb
```diff
@@ -50,9 +50,8 @@ def then_i_should_see_case_account_balance
 
   def then_i_should_see_balance_graph
     expect(page.body).to have_css('h2', text: 'Balance over time', count: 1)
+    expect(page.body).to have_css('span', text: 'Graph', count: 1)
```
kjdchapman: Does this produce a false positive? I'm not sure if this component being hidden will stop capybara finding it.

app/views/tenancies/show.html.erb
```diff
@@ -73,13 +73,20 @@
...
+      <details class="govuk-details" data-module="govuk-details">
+      <summary style="color:#1d70b8; cursor:pointer" class="govuk-details__summary">
+    <span class="govuk-details__summary-text"><u>Graph</u></span>
+  </summary>
+    <canvas id="balance_chart"></canvas> 
+    </details>
   </div>
 </div>
 
+
```
kjdchapman: Can lose these extra empty lines?

app/views/tenancies/show.html.erb
```diff
@@ -195,6 +202,7 @@
       var transactions = <%= from_last_year_as_json(@tenancy.transactions) %>
         window.arrearsGraph && window.arrearsGraph.destroy();
       window.arrearsGraph = BalanceGraph(ctx, transactions)
+      ctx.style.height = "300px"
```
kjdchapman: What does this line do?

I'm worried that a single custom style line in the markup file will make this hard to maintain in future, especially if we're trying to make these pages work on tablets at some point.

elena-vi: is there any reason why we're specifying this height here?

Miles-Alford: if we don't we get an issue where the graph is drawn with a starting height of 50px.

spec/features/view_a_case_spec.rb
maysakanoni: I think the span element is always there which is what they're testing for.

app/views/tenancies/show.html.erb
Miles-Alford: If its style is not set when it is draw, the graph starts with a height of 50px. squashing the graph similar to the issue we had when using javascript to show the graph

spec/features/view_a_case_spec.rb
kjdchapman: We talked a little around this irl about the trade-offs between introducing javascript testing to cover this feature, I was happy with testing the js-less version (where the graph will still be visible).

PR #208, opened 2019-10-03
Improve stubbing in "view a case" spec

kjdchapman:
Noticed some funky webmock/stubbing constants in these specs as I was working on MA-35 ("Link period summary/transactions to arrears diary").

This PR: 
- explictly webmocks each of the APIs for viewing a single case
- removes any stubbing of constants from this spec
- fixes a few other specs that I believe were "accidentally" passing because of this spec's heavy use of wide-matching regexes in webmock

### DIFF
--- a/spec/controllers/tenancies_controller_spec.rb
+++ b/spec/controllers/tenancies_controller_spec.rb
@@ -83,16 +83,6 @@ describe TenanciesController do
     end
   end
 
-  context '#show' do
-    it 'should assign a valid tenancy' do
-      get :show, params: { id: '1234567' }
-
-      expect(assigns(:tenancy)).to be_present
-      expect(assigns(:tenancy)).to be_instance_of(Hackney::Income::Domain::Tenancy)
-      expect(assigns(:tenancy)).to be_valid
-    end
-  end
-
   context '#pause' do
     it 'should assign a valid tenancy' do
       get :pause, params: { id: '1234567' }
--- a/spec/features/view_a_case_spec.rb
+++ b/spec/features/view_a_case_spec.rb
@@ -144,7 +144,7 @@ describe 'Viewing A Single Case' do
 
   def stub_income_api_tenancy
     response_json = File.read(Rails.root.join('spec', 'examples', 'single_case_response.json'))
-    stub_request(:get, "https://example.com:80/api/v1/tenancies/1234567%2F01")
+    stub_request(:get, "https://example.com/api/v1/tenancies/1234567%2F01")
       .with(headers: { 'X-Api-Key' => ENV['INCOME_COLLECTION_API_KEY'] })
       .to_return(status: 200, body: response_json)
   end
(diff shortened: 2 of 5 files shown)


spec/features/view_a_case_spec.rb
```diff
@@ -145,24 +142,73 @@ def then_i_should_see_a_pause_button
     expect(page).to have_link(href: '/tenancies/1234567%2F01/pause')
   end
 
-  def stub_my_cases_response
-    stub_const('Hackney::Income::IncomeApiUsersGateway', Hackney::Income::StubIncomeApiUsersGateway)
+  def stub_income_api_tenancy
+    response_json = File.read(Rails.root.join('spec', 'examples', 'single_case_response.json'))
+    stub_request(:get, 'https://example.com/tenancy/api/v1/tenancies/1234567%2F01')
```
richrace: I suggest adding a new line between `146-147` as there is between `request_json` and `stub_request` in other methods.

spec/features/view_a_case_spec.rb
```diff
@@ -145,24 +142,73 @@ def then_i_should_see_a_pause_button
...
+    response_json = {
+      data: {
+        contacts: [{
+          post_code: 'E8 1DY',
+          responsible: true,
+          address_line1: 'Primary Street',
+          title: 'Mr',
+          first_name: 'Alan',
+          last_name: 'Sugar',
+          email_address: 'alan.sugar@domain.com'
```
richrace: Should we use `alan.sugar@example.com` rather than `@domain.com`?

As `@example.com` is a special domain.

spec/controllers/tenancies_controller_spec.rb
```diff
@@ -83,16 +83,6 @@
...
 
-  context '#show' do
-    it 'should assign a valid tenancy' do
-      get :show, params: { id: '1234567' }
-
-      expect(assigns(:tenancy)).to be_present
-      expect(assigns(:tenancy)).to be_instance_of(Hackney::Income::Domain::Tenancy)
-      expect(assigns(:tenancy)).to be_valid
-    end
-  end
```
richrace: How is deleting this test related to this MR?

spec/features/view_a_case_spec.rb
kjdchapman: huh, I never knew about [this](https://www.iana.org/domains/reserved)!

spec/controllers/tenancies_controller_spec.rb
kjdchapman: Some of the `stub_requests` I changed in this PR used wide-matching regexes. I believe they were causing these (unrelated) specs to pass.

I could have fixed them, but since they cover the same functionality, I decided to just delete the controller specs (which, as I understand it, [are on the way out anyway](https://medium.com/just-tech/rspec-controller-or-request-specs-d93ef563ef11)).

PR #209, opened 2019-10-04
Add scrollable full history to tenancy details

richrace:
## Old

image

## New

image


app/controllers/tenancies_controller.rb
```diff
@@ -19,7 +19,10 @@ def index
 
   def show
     @page_number = page_number
-    @tenancy = use_cases.view_tenancy.execute(tenancy_ref: params.fetch(:id))
+
+    tenancy_ref = params.fetch(:id)
+    @tenancy = use_cases.view_tenancy.execute(tenancy_ref: tenancy_ref)
+    @actions = use_cases.view_actions.execute(tenancy_ref: tenancy_ref)
```
kjdchapman: I'm not happy about this proliferation in the controller, but to address it otherwise would take a huge amount of effort. 

I made some progress unpicking this stack of use cases/gateways, but I would like to build up more understanding before changing them significantly.

app/views/tenancies/show.html.erb
```diff
@@ -123,67 +123,37 @@
...
-<!--          <td title="<%#= action.fetch(:scheduled_for) %>">-->
-            <%#= action.fetch(:scheduled_for).strftime('%B %e %Y') %>
-<!--          </td>-->
-<!--        </tr>-->
-        <%# end %>
-<!--      </tbody>-->
-<!--    </table>-->
-    <%# end %>
-<!--  </div>-->
-<!--</div>-->
```
richrace: 💥 🎉

PR #211, opened 2019-10-07
Display the number of bedrooms

georgeschena:
WHAT: Display the number of bedrooms for a given property

WHY: So an Income Collection Officer, can see the number of bedrooms in a property

Screenshot 2019-10-07 at 15 29 19

### DIFF
--- a/lib/hackney/income/domain/tenancy.rb
+++ b/lib/hackney/income/domain/tenancy.rb
@@ -40,7 +40,7 @@ module Hackney
         end
 
         def display_number_of_bedrooms
-          return "Unknown" if case_priority[:num_bedrooms].nil?
+          return 'Unknown' if case_priority[:num_bedrooms].nil?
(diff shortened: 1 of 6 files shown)


lib/hackney/income/domain/tenancy.rb
```diff
@@ -39,6 +39,11 @@ def display_nosp_expiry
           format_date(case_priority[:nosp_expiry_date])
         end
 
+        def display_number_of_bedrooms
+          return 'Unknown' if case_priority[:num_bedrooms].nil?
```
kjdchapman: Is this value ever an empty string rather than nil? If so could use [`.blank?`](https://apidock.com/rails/Object/blank%3F) here. If not, all good.

georgeschena: Yeah the DB has uses NULL, so this is fine

PR #215, opened 2019-10-15
allowing users to filter their worktray by patch

elena-vi:
Users need the ability to filter their worktray by arrears patch. 

Screenshot 2019-10-22 at 10 58 07


elena-vi:
https://hackney.atlassian.net/secure/RapidBoard.jspa?rapidView=22&projectKey=MA&modal=detail&selectedIssue=MA-183

spec/features/view_my_cases_spec.rb
```diff
@@ -70,13 +84,19 @@ def stub_my_cases_response
...
   end
 
   def stub_my_paused_cases_response
-    stub_request(:get, /my-cases\?is_paused=true&number_per_page=20&page_number=1&user_id=/)
+    stub_request(:get, /my-cases\?is_paused=true&number_per_page=20&page_number=1&patch&user_id=/)
+      .with(headers: { 'X-Api-Key' => ENV['INCOME_COLLECTION_API_KEY'] })
+      .to_return(status: 200, body: SINGLE_CASE_RESPONCE)
+  end
+
+  def stub_my_filtered_cases_response
```
kjdchapman: These stub methods are generally in the form `stub_endpoint_name_response`, but now they've started to drift to other names. Is this confusing? 

I would like to use [shared contexts](https://relishapp.com/rspec/rspec-core/docs/example-groups/shared-context) for these stubs, I think - `with_context 'a stubbed my-cases endpoint'` feels about the most descriptive way to do this.

lib/hackney/income/patch_codes.rb
```diff
@@ -0,0 +1,35 @@
+module Hackney
+  module Income
+    class PatchCodes
```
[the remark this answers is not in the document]
richrace: Adding my thoughts.

I think that a file that looks like this, is only for the view files. So it should be a view helper a la https://github.com/LBHackney-IT/LBH-IncomeCollection/blob/def3c23384b697286aa7a2e5c11c832335672f8e/app/helpers/recommended_actions_select_helper.rb

If we want to move the data somewhere (like a Gateway), I think that might improve boundaries. But don't have a strong opinion on this part.

app/controllers/tenancies_controller.rb
```diff
@@ -58,6 +60,17 @@ def valid_tenancies(tenancies)
   end
 
   def user_assigned_cases_params
-    params.permit(:page, :recommended_actions, :paused, :full_patch, :upcoming_evictions, :upcoming_court_dates)
+    permitted_params = params.permit(:page, :recommended_actions, :paused, :full_patch, :upcoming_evictions, :upcoming_court_dates, :patch)
+
+    permitted_params[:patch] ||= cookies[:patch_code] if cookies[:patch_code].present?
```
kjdchapman: I think this has the same (or similar enough) behaviour without the `if`?

kjdchapman: Tried it, it fails some tests. Those tests might be a little overzealous, but not sure it's worth looking at for the sake of cutting down one line.

app/controllers/tenancies_controller.rb
```diff
@@ -58,6 +60,17 @@ def valid_tenancies(tenancies)
...
+
+    permitted_params
+  end
+
+  def set_patch_code_cookie
+    patch_code_param = params.permit(:patch)
+    return if patch_code_param.blank?
```
kjdchapman: As a point of interest, because I saw you talking about `if` v `unless` in another thread, this is somewhere I find the `unless` version readable:
`return unless patch_code_param.present?`

This isn't a recommendation to change, just a thought.

spec/controllers/tenancies_controller_spec.rb
```diff
@@ -85,6 +85,21 @@
         expect(assigns(:user_assigned_tenancies)).to all(be_valid)
       end
     end
+
+    context 'when filtering by patch' do
```
kjdchapman: Are you confident this covers the functionality, given it's a controller spec?

If you're happy adding to the controller specs for now, I'm happy.

spec/features/view_my_cases_spec.rb
```diff
@@ -7,8 +7,10 @@
 
   before do
     stub_my_cases_response
-    stub_my_paused_cases_response
-    stub_my_cases_filtered_on_recommended_actions_response
+    stub_my_cases_response(is_paused: true)
+    stub_my_cases_response(recommended_actions: 'no_action')
+    stub_my_cases_response(patch: 'E01')
+    stub_my_cases_response(is_paused: true, patch: 'E01')
```
kjdchapman: Nice one, this reads better to me.

spec/features/view_my_cases_spec.rb
```diff
@@ -34,6 +36,15 @@
     then_i_should_filter_worktray_by_an_action
   end
 
+  scenario do
+    given_i_am_logged_in
+    when_i_visit_the_homepage
+    i_should_see_all_of_the_tabs
```
kjdchapman: I think "should" is noise here (and in many spec descriptions). I'd write this as
`then_I_see_all_the_tabs`

(We might not keep these feature specs in this style, but I feel the same way about `it` block strings, so thought I'd bring it up now)

spec/features/view_my_cases_spec.rb
```diff
@@ -84,69 +101,31 @@ def then_i_should_filter_worktray_by_an_action
...
   end
 
-  def stub_my_cases_filtered_on_recommended_actions_response
+  def stub_my_cases_response(override_params = {})
     stub_const('Hackney::Income::IncomeApiUsersGateway', Hackney::Income::StubIncomeApiUsersGateway)
 
     response_json = File.read(Rails.root.join('spec', 'examples', 'my_cases_response.json'))
-    stub_request(:get, /my-cases\?full_patch=false&is_paused=false&number_per_page=20&page_number=1&recommended_actions=no_action&upcoming_court_dates=false&upcoming_evictions=false&user_id=/)
+
+    default_actions = {
```
kjdchapman: My reading of this is that it's `default_filters` rather than `actions`?

app/controllers/tenancies_controller.rb
richrace: https://github.com/rubocop-hq/rubocop-rails/blob/master/lib/rubocop/cop/rails/blank.rb

spec/features/view_my_cases_spec.rb
richrace: I agree. But I feel that is changing a lot of tests to ensure consistency.

spec/controllers/tenancies_controller_spec.rb
richrace: The `cookies` functionality is mainly covered by the Feature spec.

app/controllers/tenancies_controller.rb
richrace: I could rewrite it:

```suggestion
    if cookies[:patch_code].present? && permitted_params[:patch].blank?
      permitted_params[:patch] = cookies[:patch_code] 
    end
```

richrace:
> If the layout has changed, would you mind adding another screenshot?
> 
> There's `patch` in some places, `patch_code` in others. Is that potentially confusing?
> 
> I've got thoughts, but nothing that blocks this'un. :shipit: once you're happy!

The API exists `patch` so would have to update that too.

spec/features/view_my_cases_spec.rb
richrace: After this is merged in we could look at extracting this into a `spec/helper`

kjdchapman:
> What other screenshot would you want? Before/After?

I thought you'd wanted to change the layout, so the original screenshot would now be out-of-date. 

If the layout hasn't changed, I have the same concern as above - that the "Filter" button is far away from the dropdown. Happy for you to take the call on that, though.

> The API expects patch so would have to update that too (I prefer patch_code over patch)

I agree with `patch_code` over `patch`, and I like the changes made in https://github.com/LBHackney-IT/LBH-IncomeCollection/pull/215/commits/2621271905e19454d0d725546d82ce7024a459c7 - we might change the API later.

PR #216, opened 2019-10-15
Add classification to the worktray table

richrace:
### What

- Add Classification for next recommended action

### Why

- So Case Workers have help knowing what to do next

### Guidance for review

- This commit https://github.com/LBHackney-IT/LBH-IncomeCollection/pull/216/commits/dcf985074377706f7e904e206ecdd0096be0bee6 contains a large refactor of filter params to be an object.


app/controllers/tenancies_controller.rb
```diff
@@ -56,40 +52,12 @@ def update
...
-    ActiveModel::Type::Boolean.new.cast(params.fetch(:upcoming_court_dates, false))
-  end
-
-  def upcoming_evictions?
-    ActiveModel::Type::Boolean.new.cast(params.fetch(:upcoming_evictions, false))
-  end
-
-  def cases_per_page
-    20
+  def list_user_assigned_cases_params
```
kjdchapman: I'd call this `user_assigned_cases_params` - I don't think the first word adds understanding.

app/views/tenancies/_worktray_table.html.erb
```diff
@@ -39,6 +40,14 @@
           <span class="tenancy_list__ref"><%= human_agreement_status(tenancy.current_arrears_agreement_status) %></span>
         <% end %>
       </td>
+
+      <td>
+        <%= link_to(tenancy_path(id: tenancy.ref, page_params: @page_params), class: 'tenancy_list__link') do %>
+          <span>
+            <%= tenancy.classification ? recommended_actions_map[tenancy.classification.to_sym] : '' %>
```
kjdchapman: if classifications are strings, is it better to have the `recommended_actions_map` have strings as keys? The `to_sym` call feels a little awkward.

I'm not sure if there's a performance/other implication here.

app/views/tenancies/index.html.erb
```diff
@@ -1,23 +1,35 @@
...
-  <%= worktray_tab_link_to("Active", worktray_path) %>
+  <%= worktray_tab_link_to("Immediate Actions", worktray_path) %>
   <%= worktray_tab_link_to("Paused", worktray_path(paused: true), :paused) %>
-  <%# <%= worktray_tab_link_to("Full Patch", worktray_path(full_patch: true), :full_patch) %>
+  <%= worktray_tab_link_to("Full Patch", worktray_path(full_patch: true), :full_patch) %>
   <%# <%= worktray_tab_link_to("Upcoming Court Dates", worktray_path(upcoming_court_dates: true), :upcoming_court_dates) %>
   <%# <%= worktray_tab_link_to("Upcoming Evictions", worktray_path(upcoming_evictions: true), :upcoming_evictions) %>
 
+<table>
+   <tr>
```
kjdchapman: nit: The indentation looks wrong to me, here - three spaces instead of two?

app/views/tenancies/index.html.erb
```diff
@@ -1,23 +1,35 @@
...
+     <%= form_tag(worktray_path, :method => "get") do %>
+       <td>
+          <label class="govuk-label" for="code"><strong>Next Recommended Action</strong><br/></label>
```
kjdchapman: I would use the [govuk design system typography override class](https://design-system.service.gov.uk/styles/typography/#font-override-classes) `govuk-!-font-weight-bold` on the label here rather than a strong tag.

The emphasis is more for visual design than semantic, I think the class is more readable/expressive.

app/views/tenancies/index.html.erb
```diff
@@ -1,23 +1,35 @@
...
+       <td>
+         <%= submit_tag('Filter by next action', class: 'button') %>
+       </td>
+     <% end %>
+   </tr>
+ </table>
+
   <% if @user_assigned_tenancies.none? %>
-    <h3 class="tenancy_list tenancy_list__no_tenancies">You currently have no assigned cases</h3>
+    <h3 class="tenancy_list tenancy_list__no_tenancies">No results found</h3>
```
kjdchapman: This wording strikes me as too "technical", but this is very much an opinion.
> There are no cases for the filters you have chosen.

(in fact, ideally)
> There are no cases for [human-readable expression of your filter]

e.g.
> There are no cases assigned to you.
> There are no paused cases assigned to you.
> There are no cases that need immediate action.

That second option is probably more work, though.

app/views/tenancies/index.html.erb
```diff
@@ -1,23 +1,35 @@
...
 
     <div class='other_actions tenancy_list column-align-center'>
       <%= link_to 'Send letters', letters_new_path, class: 'button' %>
       <%= link_to  'View letters', documents_path, class: 'button' %>
     </div>
+  <% else %>
```
kjdchapman: nice work cleaning this conditional up :+1:

lib/hackney/income/filter_params/list_user_assigned_cases_params.rb
```diff
@@ -0,0 +1,49 @@
...
+          @immediate_action     = options[:immediate_action]
+          @recommended_actions  = options[:recommended_actions]
+          @full_patch           = cast_boolean(options[:full_patch])
+          @upcoming_court_dates = cast_boolean(options[:upcoming_court_dates])
+          @upcoming_evictions   = cast_boolean(options[:upcoming_evictions])
+          @paused               = cast_boolean(options[:paused])
+        end
+
+        def count_per_page
+          @count_per_page&.to_i || 20
```
kjdchapman: On the fence as to whether these are unacceptable magic numbers or if extracting them would be overkill.

spec/features/view_my_cases_spec.rb
```diff
@@ -26,6 +27,13 @@
     then_i_should_see_paused_cases
   end
 
+  scenario do
+    given_i_am_logged_in
+    when_i_visit_the_homepage
+    i_should_see_all_of_the_tabs
+    then_i_should_filter_worktray_by_an_action
```
kjdchapman: This doesn't read like something a user would understand, to me. Is the last part meant to be "I should see the worktray filtered by an action"?

Here's the scenarios (with my naive understanding of this feature) i might expect:
> Given I'm logged in
> When I visit the homepage
> Then I'll see ways to filter 

> Given I'm logged in
> And I'm looking at the home page
> When I select <some filter>
> Then I'll only see cases for that filter

spec/features/view_my_cases_spec.rb
```diff
@@ -26,6 +27,13 @@
...
+  end
```
kjdchapman: I'm not sure if we should be creating more scenario-style specs, rather than request specs. They're not striking me as particularly readable/useful so far.

spec/features/view_my_cases_spec.rb
```diff
@@ -79,6 +93,15 @@ def stub_my_cases_response
       .to_return(status: 200, body: response_json)
   end
 
+  def stub_my_cases_filtered_on_recommended_actions_response
```
kjdchapman: I don't like the proliferation of these `stub_endpoint_with_some_conditions` methods. 

I think they're risking confusion over what the test setup is (because we're adding multiple stubs for the same endpoint). 

I'm not sure what the best alternative is. The first one that comes to mind is parameterizing these methods, but that might not be it either. It might be that the stubs are sunk further into the tests alongside context? I'd have to dive into this a bit to offer more concrete advice, I think.

app/views/tenancies/_worktray_table.html.erb
richrace: I like the pretty symbol hash in the view helper and the time taken to go from string to symbol  I think is worth it.

Can always have the Tenancy List Item Domain object return a symbol if this line bothers you?

richrace:
> This is good stuff, and supports the future work well.
> 
> I am worried that there's now a component that both a Gateway and Use Case rely on. I think this change doesn't fix the "bad" dependency flow because of that.
> 
> Maybe we should get round a whiteboard at some point to get a shared understanding of use cases / gateways?

I agree that we've just moved code from one place to another. 

We could do the `filter_params.to_params` (with a better method name) in the UseCase constructor to then mean we only have a simple hash that gets passed around rather than a "full blown object"?

spec/features/view_my_cases_spec.rb
richrace: Yea, not happy with the amount of stubs here. But I think it's a symptom of having one test file for many tests. 

Or, we add the specific test setup in each `scenario`?

lib/hackney/income/filter_params/list_user_assigned_cases_params.rb
richrace: Could you please explain what "unacceptable" means here?

Would you move `20` to be a constant somewhere? If so where? I think it's fine here are these are the default params.

app/views/tenancies/index.html.erb
georgeschena: This worked `form-label-bold` this didn't `govuk-!-font-weight-bold`

PR #217, opened 2019-10-21
Show filter only on immediate action tab

georgeschena:
### WHAT: 
Only show immediate action filter on immediate actions tab

### WHY: 
As users will only need to use this action on this tab only

### DIFF
--- a/app/views/tenancies/index.html.erb
+++ b/app/views/tenancies/index.html.erb
@@ -7,9 +7,7 @@
   <%# <%= worktray_tab_link_to("Upcoming Court Dates", worktray_path(upcoming_court_dates: true), :upcoming_court_dates) %>
   <%# <%= worktray_tab_link_to("Upcoming Evictions", worktray_path(upcoming_evictions: true), :upcoming_evictions) %>
 
-  <% unless %w[paused full_patch upcoming_court_dates upcoming_evictions].any? { |tab| params.include? tab } %>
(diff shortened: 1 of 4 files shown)


app/views/tenancies/index.html.erb
```diff
@@ -7,19 +7,9 @@
...
-          <label class="govuk-label form-label-bold" for="code">Next Recommended Action<br/></label>
-          <%= select_tag(:recommended_actions, recommended_actions_dropdown_options(selected: params[:recommended_actions]), { :class => 'form-control' }) %>
-        </td>
-        <td>
-          <%= submit_tag('Filter by next action', class: 'button') %>
-        </td>
-      <% end %>
-    </tr>
-  </table>
+  <% unless %w[paused full_patch upcoming_court_dates upcoming_evictions].any? { |tab| params.include? tab } %>
```
richrace: I try to avoid `unless` statements. I generally find them quite hard to understand. They are my last resort or if I have to get something out quickly. So I would always try to convert them to be `if ...`.

I'd like this to be a descriptive method. That way I can see what/why the array of `%w[paused full_patch upcoming_court_dates upcoming_evictions]` is important.

Having it as a method would allow me to write the render like this:

```ruby 
  <%= render 'tenancies/worktray/recommended_action_filter' if something_is_true? %>
```

Which I find easy to understand.

I also find `.any? { |tab| params.include? tab }` slightly long winded and not very clear.

PR #221, opened 2019-10-24
Hide 'send' buttons for letter before action

kjdchapman:
Notably, this PR doesn't extract this policy decision - _don't use notify for certain types of letter_ - into its own component.

(My hope is we'll make that policy more visible as we work on getting the 'Download' functionality working)

If you select LBA, you'll see (with fake content):

image

### DIFF
--- a/app/views/letters/preview.erb
+++ b/app/views/letters/preview.erb
@@ -59,9 +59,11 @@
         </div>
       </div>
 
-      <div>
-        <%= button_tag "Confirm and Send All", class: "button", onclick: 'submit_send_all_letters(event)' %>
-      </div>
+      <% if @preview[:sendable] %>
+        <div>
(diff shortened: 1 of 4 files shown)


app/views/letters/preview.erb
```diff
@@ -59,9 +59,11 @@
         </div>
       </div>
 
-      <div>
-        <%= button_tag "Confirm and Send All", class: "button", onclick: 'submit_send_all_letters(event)' %>
-      </div>
+      <% if @preview[:sendable] %>
+        <div>
+          <%= button_tag "Confirm and Send All", class: "button", onclick: 'submit_send_all_letters(event)' %>
```
richrace: Note: Not a fan off JS `onclick` here 😱

PR #222, opened 2019-10-28
Add privacy notice to welcome page

alex-demetriou:
### Why
Users will be advised that information about their usage will be sent to Google Analytics and Hotjar when logging in.

### What
[MA-210](https://hackney.atlassian.net/browse/MA-210): Add Privacy Text to the signing in page

### DIFF
--- a/app/views/sessions/new.html.erb
+++ b/app/views/sessions/new.html.erb
@@ -8,11 +8,12 @@
     <%= link_to auth_provider_path do %>
       <%= image_tag 'microsoft_login.png', alt: 'Sign in with Microsoft' %>
     <% end %>
-    
-    <br><br><br>
+
+    <div class="privacy-notice">
       <h3>Privacy Notice</h3>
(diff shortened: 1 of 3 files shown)


kjdchapman:
This looks sensible to me, but can you include a screenshot of how it looks on the page?

alex-demetriou:
> This looks sensible to me, but can you include a screenshot of how it looks on the page?

Added

app/views/sessions/new.html.erb
```diff
@@ -8,5 +8,11 @@
     <%= link_to auth_provider_path do %>
       <%= image_tag 'microsoft_login.png', alt: 'Sign in with Microsoft' %>
     <% end %>
+    
+    <br><br><br>
```
richrace: I'm not too keen on this. I'd rather see the privacy in a `div` and some CSS applied

@kjdchapman thoughts?

kjdchapman: Yes, I agree (I feel fairly strongly about it too!)

PR #224, opened 2019-10-30
Add Download button for Letters

richrace:
### Why

- Allow users to be able to download PDFs 

### What

- Add a download button


kjdchapman:
Ah, I hadn't twigged that was one of the advantages of these. I'm happy for us to add them as long as they're useful!

PR #229, opened 2019-11-05
Use Google Auth and remove user

richrace:
### What

- Remove Azure AD
- Add Google Auth
- Remove references to API User models


app/controllers/application_controller.rb
```diff
@@ -9,46 +11,61 @@ class ApplicationController < ActionController::Base
...
 
   def check_authentication
-    return if logged_in? || auth_request? || logout_request?
+    return if logged_in?
 
     redirect_to login_path
   end
 
+  def read_hackney_token
+    @current_user if @current_user.present?
```
bjpirt: Trying to work out if I'm missing something, but I don't think this does anything does it?

richrace: 🤦‍♂

richrace:
Need to update this action: https://github.com/LBHackney-IT/LBH-IncomeCollection/blob/use-google-auth-and-remove-user/app/controllers/tenancies_controller.rb#L40

app/views/tenancies/index.html.erb
```diff
@@ -3,38 +3,10 @@
...
-  <%= render 'tenancies/worktray/recommended_action_filter' if show_immediate_actions_filter? %>
-
-  <% if @tenancies.none? %>
-    <h3 class="tenancy_list tenancy_list__no_tenancies">No results found</h3>
-
-    <div class='other_actions tenancy_list column-align-center'>
-      <%= link_to 'Send letters', letters_new_path, class: 'button' %>
-      <%= link_to  'View letters', documents_path, class: 'button' %>
-    </div>
+  <% if current_user.leasehold_services? %>
```
richrace: This might need to change to be `if current_user.income_collection?` due to it being the only group.

This needs to be double checked before deployment.

lib/hackney/income/stub_tenancy_gateway_builder.rb
```diff
@@ -34,9 +34,8 @@ def initialize(api_host: nil, api_key: nil, include_developer_data: nil)
               @tenancies = default_tenancies
             end
 
-            def get_tenancies(user_id:, filter_params:)
+            def get_tenancies(filter_params:)
               cases = @tenancies
-                .select { |t| t.fetch(:assigned_user_id) == user_id }
                 .map(&method(:create_tenancy_list_item))
```
kjdchapman: I know we're not changing this line, but whilst we're here, I've not seen this pattern before. Any ideas?

I'd expect this to just be a one-liner: 
`cases = @tenancies.map(&:create_tenancy_list_item)`

spec/controllers/application_controller_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+            'sub' => '100518888746922116647',
+            'email' => 'hackney.user@test.hackney.gov.uk',
+            'iss' => 'Hackney',
+            'name' => 'Hackney User',
+            'groups' => ['bogus_group'],
+            'iat' => 1_570_462_732
+          }
+        end
+
+        xit 'redirects to login page' do
```
kjdchapman: Is this assertion ignored accidentally?

spec/controllers/application_controller_spec.rb
```diff
@@ -0,0 +1,109 @@
+require 'rails_helper'
+
+describe ApplicationController, type: :controller do
```
kjdchapman: I've found controller specs like this to be brittle and difficult to read - is there a reason to use them over request specs for this functionality?

spec/controllers/documents_controller_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 describe DocumentsController do
   before do
-    stub_authentication
+    sign_in
```
kjdchapman: This (and the `let(:user)` block below) is repeated in specs a lot - could it be a `shared_context`?

(I'd probably not do it as part of this large PR, thinking for future work)

spec/controllers/application_controller_spec.rb
kjdchapman: I note we've changed a lot of existing controller specs to get them to work - I'm definitely not asking us to port all of them now! I'm specifically asking about this brand-new spec.

spec/features/creating_action_diary_entry_spec.rb
```diff
@@ -1,34 +1,29 @@
...
-  end
-
   context 'filling in the form as a user' do
     it 'should display the form and call the usecase ' do
       expect(create_action_diary_entry_double).to receive(:execute).with(
         tenancy_ref: '1234567',
         action_code: 'DEB',
         comment: 'Test comment.',
-        user_id: stub_user_id
+        username: 'Hackney User'
```
kjdchapman: What the 'right' username to assert against is implicit in `create_jwt_token`. Someone might have trouble tracking it down later. Is there a good way to make it explicit here?

(My first thought is to parameterize `create_jwt_token`, so I can explicitly call it with `username: foo`, but have a default for specs that don't care)

spec/features/letters/view_letter_preview_success_spec.rb
```diff
@@ -1,12 +1,14 @@
...
 
 describe 'Viewing A Letter Preview' do
-  around { |example| with_mock_authentication { example.run } }
-
   let(:uuid) { SecureRandom.uuid }
   let(:preview) { Faker::DumbAndDumber.quote }
   let(:document_id) { Faker::Number.between(1, 1_000) }
 
+  before do
+    create_jwt_token
```
kjdchapman: In this style of spec, I would have put this setup step in `given_i_am_logged_in` - I think it's more expressive.

spec/lib/hackney/income/domain/user_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+  before do
+    user.groups = groups
+  end
+
+  describe '#leasehold_services?' do
+    context 'when there are no groups' do
+      let(:groups) { [] }
+
+      it 'returns false' do
+        expect(user.leasehold_services?).to eq(false)
```
kjdchapman: nit: I think it's more idomatic to test these are `truthy`/`falsey` rather than `true`/`false`, might sometimes reveal simpler implementations too.

spec/request/hackney_auth_session_controller_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'rails_helper'
+
+describe HackneyAuthSessionController, type: :request do
+  it 'renders the correct Auth URL' do
+    get login_path
+
+    expect(response).to be_successful
+    expect(response).to render_template(:show)
```
kjdchapman: Does this throw a deprecation warning? I think checking something renders a certain template will be brittle/hard to read.

kjdchapman: Same thought for the new `worktray_spec`.

spec/support/mock_auth_helper.rb
```diff
@@ -1,27 +1,40 @@
...
-      'name' => Faker::StarTrek.character,
-      'email' => authable_email,
-      'first_name' => Faker::StarTrek.specie,
-      'last_name' => Faker::StarTrek.villain
-    },
-      'extra' => {
-      'raw_info' => {
-        'id_token' => "#{Faker::Number.number(6)}.123456ABC"
-      }
+module MockAuthHelper
```
kjdchapman: I'm not sure why there are two methods here. 

When should I use `sign_in` and when should I `create_jwt_token`?

app/controllers/application_controller.rb
```diff
@@ -9,46 +11,57 @@ class ApplicationController < ActionController::Base
...
   private
 
   def check_authentication
-    return if logged_in? || auth_request? || logout_request?
+    return if logged_in?
 
     redirect_to login_path
   end
 
+  def read_hackney_token
```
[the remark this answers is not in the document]
richrace: Could be moved into a UseCase in a tidy up phase.

app/controllers/application_controller.rb
```diff
@@ -9,46 +11,57 @@ class ApplicationController < ActionController::Base
...
+      raw_hackney_token, ENV['HACKNEY_JWT_SECRET'], true, algorithm: 'HS256'
+    ).first
+
+    @current_user = Hackney::Income::Domain::User.new.tap do |u|
+      u.id = payload['sub']
+      u.name = payload['name']
+      u.email = payload['email']
+      u.groups = payload['groups']
+    end
+  rescue JWT::DecodeError => e
```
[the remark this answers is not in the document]
richrace: This was more around being defensive.

app/controllers/dev_session_controller.rb
```diff
@@ -0,0 +1,22 @@
+# Don't include this file in anything but Development as it auto-logs you in!
+return unless Rails.env.development?
```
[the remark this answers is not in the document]
richrace: The main motivation for using `Rails.env.development?` is we do not want the ability to use this outside of dev.

app/controllers/dev_session_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+      'groups' => ['group 1', 'group 2'],
+      'iat' => 1_570_462_732
+    }
+
+    jwt_token = JWT.encode(jwt_payload, ENV['HACKNEY_JWT_SECRET'], 'HS256')
+    cookies['hackneyToken'] = jwt_token
+
+    redirect_to :root
+  end
+end
```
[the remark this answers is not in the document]
richrace: Similar to the comment above, this file would not run under tests as it would not be included in the `Rails.env.test` environment

app/controllers/hackney_auth_session_controller.rb
```diff
@@ -0,0 +1,13 @@
+class HackneyAuthSessionController < ApplicationController
+  skip_before_action :check_authentication
+
+  def show
```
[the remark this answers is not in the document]
richrace: Rails has the concepts of `resources` and `resource`.

Resources -> Has many things -> `/sessions/:id -> SessionsController#show`
Resource -> Is a singular thing -> `/session -> SessionsController#show` 

We are not using the Rails magic route helpers for the routes as we're following the existing route structure.

https://guides.rubyonrails.org/routing.html#crud-verbs-and-actions
https://guides.rubyonrails.org/routing.html#singular-resources

app/controllers/letters_controller.rb
```diff
@@ -22,11 +22,7 @@ def preview
   end
 
   def ajax_preview
-    @preview = use_cases.get_letter_preview.execute(
-      template_id: params.require(:template_id),
-      pay_ref: params.require(:pay_ref),
-      user_id: session[:current_user].fetch('id')
-    )
+    @preview = generate_letter_preview(params.require(:pay_ref))
```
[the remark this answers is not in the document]
richrace: This was a DRY change: https://github.com/LBHackney-IT/LBH-IncomeCollection/blob/master/app/controllers/letters_controller.rb#L63-L69

app/controllers/tenancies_controller.rb
```diff
@@ -6,23 +6,22 @@ class TenanciesController < ApplicationController
...
     @page_params = request.query_parameters
 
-    @tenancies = Kaminari.paginate_array(@user_assigned_tenancies).page(@page_number)
+    @tenancies = Kaminari.paginate_array(@tenancies).page(@page_number)
   end
 
   def show
     @previous_page_params = request.query_parameters[:page_params]
-    @page_number = user_assigned_cases_params[:page]
+    @page_number = list_cases_params[:page]
```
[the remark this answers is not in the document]
richrace: I think we need to think about naming as a separate issue.

app/views/hackney_auth_session/show.html.erb
```diff
@@ -5,9 +5,7 @@
     <h2>You must sign in to continue</h2>
     <p>Click below to sign in with your <strong>hackney.gov.uk</strong> account.</p>
 
-    <%= link_to auth_provider_path do %>
-      <%= image_tag 'microsoft_login.png', alt: 'Sign in with Microsoft' %>
-    <% end %>
+    <%= link_to "Sign in with Google", Rails.application.config.x.auth_url %>
```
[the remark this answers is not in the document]
richrace: I think so, to make sure they know it will send them to Google rather than Azure.

app/views/tenancies/index.html.erb
richrace: I do not think so.

config/environments/production.rb
```diff
@@ -81,6 +81,8 @@
     config.logger    = ActiveSupport::TaggedLogging.new(logger)
   end
 
+  config.x.auth_url = ENV.fetch('AUTH_URL', 'https://auth.hackney.gov.uk/auth?redirect_uri=https://managearrears.hackney.gov.uk/')
```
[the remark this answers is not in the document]
richrace: We started off with two separate parts but then decided to do it as one. I can't remember why. 

/cc @bjpirt

config/routes.rb
```diff
@@ -26,9 +25,10 @@
...
 
-  get '/login', to: 'sessions#new', as: :login
-  get '/logout', to: 'sessions#destroy', as: :logout
-  match '/auth/:provider/callback', to: 'sessions#create', via: %i[get post]
-  match '/auth/failure', to: 'sessions#failure', via: %i[get post]
-  match '/auth/:provider/failure', to: 'sessions#failure', via: %i[get post]
+  get '/login', to: 'hackney_auth_session#show'
+  get '/logout', to: 'hackney_auth_session#destroy'
+
+  get '/login/dev', to: 'dev_session#new' if Rails.env.development?
```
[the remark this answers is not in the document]
richrace: Correct - this route will only exist in Development.

lib/hackney/income/domain/user.rb
```diff
@@ -0,0 +1,17 @@
+module Hackney
+  module Income
+    module Domain
+      class User
+        attr_accessor :id, :name, :email, :groups
+
+        def leasehold_services?
+          groups.join(' ').include?('leasehold')
```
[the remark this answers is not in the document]
richrace: Nope. Quick and dirty to get something working. There is a ticket to do this properly.

lib/hackney/income/letters_gateway.rb
```diff
@@ -31,11 +32,12 @@ def create_letter_preview(payment_ref:, template_id:, user_id:)
         JSON.parse(res.body).deep_symbolize_keys
       end
 
-      def send_letter(uuid:, user_id:)
+      def send_letter(uuid:, user:)
         uri = URI("#{@api_host}#{SEND_LETTER_ENDPOINT}")
         body_data = {
           uuid: uuid,
-          user_id: user_id
+          username: user.name,
```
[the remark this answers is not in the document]
richrace: Will confirm - need to look at the backend.

lib/hackney/income/stub_tenancy_gateway_builder.rb
richrace: `@tenancies` is an array of Hashes. 

A Hash does not have method `#create_tenancy_list_item`.

The `#create_tenancy_list_item` is within this file, so we can use the `method(...)` function to give scope access.

https://www.sethvargo.com/ruby-local-method-map-shortcut/ tells you how to do it with the sugar (they updated the article to use the sugar at the end)

spec/controllers/application_controller_spec.rb
richrace: The idea of this test was to be more of a unit test. Happy for it to be refactored away into other tests.

spec/controllers/application_controller_spec.rb
richrace: Yes - we do not have groups that deny access at the moment.

spec/controllers/documents_controller_spec.rb
richrace: Yes, there could be improvements around this.

spec/features/creating_action_diary_entry_spec.rb
richrace: Similar to an earlier comment, we probably need to think about how we want to have this setup in the tests more than "lets just get it working".

spec/lib/hackney/income/domain/user_spec.rb
richrace: I'm personally not a fan of `truthy` or `falsey` when I'm expecting a boolean as the return value as the method name implies (via the `?`)

spec/request/hackney_auth_session_controller_spec.rb
richrace: From the official docs: https://relishapp.com/rspec/rspec-rails/docs/request-specs/request-spec 🤷‍♂

spec/support/mock_auth_helper.rb
richrace: - `sign_in` is for Controller tests.
- `create_jwt_token` for Feature tests
- `build_jwt_token` required for Request tests

I agree it's not ideal.

app/views/tenancies/worktray/_tabs.html.erb
```diff
@@ -0,0 +1,28 @@
...
+      </div>
+    <% end %>
+  </div>
+</div>
+
+<%= worktray_tab_link_to("Immediate Actions", worktray_path) %>
+<%= worktray_tab_link_to("Paused", worktray_path(paused: true), :paused) %>
+<%= worktray_tab_link_to("Full Patch", worktray_path(full_patch: true), :full_patch) %>
+<%# <%= worktray_tab_link_to("Upcoming Court Dates", worktray_path(upcoming_court_dates: true), :upcoming_court_dates) %>
+<%# <%= worktray_tab_link_to("Upcoming Evictions", worktray_path(upcoming_evictions: true), :upcoming_evictions) %>
```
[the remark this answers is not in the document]
richrace: These are now uncommented as the features are in master 🎉

app/views/tenancies/index.html.erb
richrace: Updated this so if you're both you see both buttons.

lib/hackney/income/letters_gateway.rb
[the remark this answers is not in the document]
richrace: Works okay 😄

app/controllers/application_controller.rb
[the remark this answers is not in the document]
kjdchapman: I'm not sure what you mean. Want to discuss this one in person?

app/controllers/dev_session_controller.rb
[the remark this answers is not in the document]
kjdchapman: I agree with the motivation!

I think I trust this conditional less than you do, but it'd be difficult for it to go wrong. If you're happy with this approach, I'm happy.

app/controllers/dev_session_controller.rb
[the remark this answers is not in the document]
kjdchapman: I think this is in the muddy space of "do we test a framework's behaviour"?

When I have a critical piece of functionality like this, I don't like to trust a framework to implement it. I'd want an acceptance spec that breaks if, say, the framework behaviour changes. 

This feels like personal preference, though - if you're happy implementing it this way, I'm not overly worried.

spec/controllers/application_controller_spec.rb
kjdchapman: I've not had a lot of success treating controllers as units that need testing - I've had more success treating them as thin "orchestration" classes that delegate any complication.

Generally, I'll test them implicitly through acceptance specs, and if they're getting complicated, extract that complication to a separate unit, and test that.

spec/controllers/application_controller_spec.rb
kjdchapman: If I'm understanding you right, this assertion is ignored on purpose because the functionality isn't implemented yet?

Is there any way to express this in code?

app/controllers/dev_session_controller.rb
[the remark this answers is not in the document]
richrace: The argument of:
> I'd want an acceptance spec that breaks if, say, the framework behaviour changes.
is valid.

spec/support/mock_auth_helper.rb
kjdchapman: Would renaming them help to expose those differences to a fresh reader?

spec/controllers/application_controller_spec.rb
richrace: I think it stems from this: https://github.com/LBHackney-IT/LBH-IncomeCollection/pull/229#discussion_r344728197

If we move the logic then this no longer a "unit" test and then we can rely on Acceptance via Requests.

app/controllers/hackney_auth_session_controller.rb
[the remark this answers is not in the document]
kjdchapman: I think I understand the above, but this route still doesn't make sense to me. Can we talk in person?

lib/hackney/income/filter_params/list_cases_params.rb
```diff
@@ -1,7 +1,7 @@
 module Hackney
   module Income
     module FilterParams
-      class ListUserAssignedCasesParams
+      class ListCasesParams
```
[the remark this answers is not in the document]
kjdchapman: Think you're right to say we deal with this separately!

spec/lib/hackney/income/domain/user_spec.rb
kjdchapman: We talked briefly in-person about this - what I took away was "we have different beliefs about what `?` means in idomatic ruby", which is fun!

I'm fine with this as-is, was just interested to share the thought :)

spec/request/hackney_auth_session_controller_spec.rb
kjdchapman: Huh, interesting. I suspect those docs are out of date. Regardless of deprecation status, I think there's a solid reason to not use this method!

The use of `render_template` here is testing the _internal details_ of a particular feature rather than the _external behaviour_. 

That means it'll break when we change the underlying structure, rather than only when the external behaviour has changed. That's not the feedback I want from my acceptance specs!

.env.test
```diff
@@ -1,3 +1,5 @@
 INCOME_COLLECTION_API_HOST=https://example.com/tenancy/api
 INCOME_COLLECTION_LIST_API_HOST=https://example.com/income/api
 INCOME_COLLECTION_API_KEY='TEST_API_KEY'
+
+HACKNEY_JWT_SECRET='test_secret'
```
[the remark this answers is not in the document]
kjdchapman: Talked in-person about this. There's difficulty with my suggested approach right now - there's a lot of tests that would need this `before` block introduced, and we'd need to clean up the ENV changes in an `after` block somewhere too.

As we're still to address our testing strategy in general, we'll probably look at this sort of thing then, not now.

app/controllers/application_controller.rb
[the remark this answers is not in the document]
kjdchapman: We talked about this in-person - we want loud errors for us, but ideally we don't want to disrupt a user's experience.

We talked about creating a new ticket - one that'll allow us to know when our frontend is encountering exceptional behaviour, so we can respond quickly.

app/controllers/dev_session_controller.rb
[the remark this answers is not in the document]
kjdchapman: We talked about this in-person. We're aligned on wanting an acceptance spec for this behaviour, but we're not agreed on how best to write/test this. Wouldn't mind talking with @bjpirt about it!

We also talked about how to feature flag this, @richrace is keen on using Rails.env, I'm less keen on that, but I'm happy to defer to the implementer.

app/controllers/dev_session_controller.rb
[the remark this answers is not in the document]
kjdchapman: This is discussed in https://github.com/LBHackney-IT/LBH-IncomeCollection/pull/229#discussion_r345181094

app/controllers/hackney_auth_session_controller.rb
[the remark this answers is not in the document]
kjdchapman: We talked about this in-person, this can be a `/sessions/new` path (or that's what we'll try next at least!)

spec/controllers/application_controller_spec.rb
kjdchapman: Talked about this in-person - we're likely to use this imminently, so it's not a worry to leave it like this. @richrace mentioned referencing this test in the upcoming work ticket so someone can find it.

spec/support/mock_auth_helper.rb
kjdchapman: Talked about this in-person - we're happy to address this as a separate piece ("how we're logging in for tests").

spec/features/letters/view_letter_preview_success_spec.rb
kjdchapman: Same comment as https://github.com/LBHackney-IT/LBH-IncomeCollection/pull/229#discussion_r345185632

spec/request/hackney_auth_session_controller_spec.rb
kjdchapman: We talked about this in-person, and I think we cleared up our shared understanding of what the role of feature/request/controller specs are for us.

Again, this is a good one to take forward into us talking about our testing strategy in general. Let's put something in a backlog to cover that.

PR #231, opened 2019-11-06
Display eviction date

maysakanoni:
**What?**

- Display the eviction date on the case view

**Why?**

- So the caseworker has is aware of any upcoming eviction dates

### DIFF
--- a/app/views/common/_tenancy_header.html.erb
+++ b/app/views/common/_tenancy_header.html.erb
@@ -13,6 +13,8 @@
       <li><strong>NoSP expires:</strong> <%= format_date(@tenancy.nosp_expiry) %></li>
       <li><strong>Court Date:</strong> <%= format_date(@tenancy.court_date) %></li>
       <li><strong>Court Outcome:</strong> <%= court_outcome_for_code(@tenancy.court_outcome) %></li>
+      <li><strong>Court Outcome:</strong> <%= court_outcome_for_code(@tenancy.court_outcome) %></li>
+      <li><strong>Evicition Date:</strong> <%= format_date(@tenancy.eviction_date) %></li>
     </ul>
   </div>
 </div>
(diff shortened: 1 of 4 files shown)


app/views/common/_tenancy_header.html.erb
```diff
@@ -13,6 +13,8 @@
       <li><strong>NoSP expires:</strong> <%= format_date(@tenancy.nosp_expiry) %></li>
       <li><strong>Court Date:</strong> <%= format_date(@tenancy.court_date) %></li>
       <li><strong>Court Outcome:</strong> <%= court_outcome_for_code(@tenancy.court_outcome) %></li>
+      <li><strong>Court Outcome:</strong> <%= court_outcome_for_code(@tenancy.court_outcome) %></li>
```
richrace: Duplicated line

richrace:
Also need to do this too https://github.com/LBHackney-IT/LBH-IncomeCollection/pull/230/files#diff-96fba83b72905fde38fb614f4176da2cR61

app/views/common/_tenancy_header.html.erb
maysakanoni: resolved in 9a14f2931eec683227891ca27bdcbe4dec09e00b

maysakanoni:
> Also need to do this too https://github.com/LBHackney-IT/LBH-IncomeCollection/pull/230/files#diff-96fba83b72905fde38fb614f4176da2cR61

Thanks, thats done here efde746c73afd1b6f147b59255455c38d8c7970d

maysakanoni:
> Need to add a line like this https://github.com/LBHackney-IT/LBH-IncomeCollection/pull/230/files#diff-3b77351e826650335b149ea8419725e5R12 to the Patch Filtering

This is done here now, 00458cb11b76210e4ce46bb1a42659eb4e5ec74f

PR #232, opened 2019-11-08
Move details to tenancy header

georgeschena:


kjdchapman:
@georgeschena is this still a desirable change? Worried it's going stale.

georgeschena:
> @georgeschena is this still a desirable change? Worried it's going stale.

Killing it

PR #237, opened 2019-11-13
Redirect GET /letters/preview to /letters/new

alex-demetriou:
### Why
* Users see an error when attempting to refresh on a letter preview

### What
* Added route to redirect a GET on /letters/preview to /letters/new


spec/features/letters/view_letter_preview_spec.rb
```diff
@@ -0,0 +1,164 @@
...
+    end
+  end
+
+  def then_i_am_redirected_to_new_letter_page
+    expect(page).to have_current_path(url_for(action: 'new', controller: 'letters'))
+  end
+
+  def then_i_refresh_the_page
+    visit current_url
+  end
```
alex-demetriou: This is the only we could get the page to refresh.
I don't know if this *definitely always* means GET, but it seems to in this case at least.

richrace:
I'm worried about how you've merged rebased master. I should not see all those commits.

spec/features/letters/view_letter_preview_spec.rb
richrace: So I take it `driver.navigate.refresh` didn't work?

alex-demetriou: I think we did try that, but I got the impression we needed selenium.
Alternatively, I may have just got confused. Is there an important difference?

alex-demetriou:
> I'm worried about how you've merged/rebased master. I should not see all those commits.

I'm going to make a new branch since I've now done it the cherry-pick way and it's much nicer. Is that ok?

PR #240, opened 2019-11-13
Redirect GET /letters/preview to /letters/new

alex-demetriou:
### Why
* Users see an error when attempting to refresh on a letter preview

### What
* Added route to redirect a GET on /letters/preview to /letters/new

### DIFF
--- a/spec/features/letters/view_letter_preview_spec.rb
+++ b/spec/features/letters/view_letter_preview_spec.rb
@@ -78,28 +78,6 @@ describe 'Viewing A Letter Preview' do
     expect(page).to have_button('Confirm and Send All', count: 1)
   end
 
-  def then_i_cannot_send_a_letter
-    expect(page).to_not have_button('Send')
-    expect(page).to_not have_button('Confirm and Send All')
-  end
-
-  def then_there_is_a_clickable_download_button
-    expect(page).to have_link('Download', href: document_path(document_id), id: "download-letter-doc-#{document_id}")
-    find("#download-letter-doc-#{document_id}[download]")
-  end
-
-  def then_there_is_not_a_clickable_download_button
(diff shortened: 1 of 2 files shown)


spec/features/letters/view_letter_preview_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+      stub_my_cases_response
+      stub_get_templates_response
+      stub_success_post_send_letter_response
+    end
+
+    scenario do
+      given_i_am_logged_in
+      when_i_visit_new_letter_page
+      then_i_see_a_letter_form
+    end
```
richrace: This also exists here https://github.com/LBHackney-IT/LBH-IncomeCollection/blob/master/spec/features/letters/view_letter_preview_success_spec.rb#L19-L23

This should only be tested in one place.

spec/features/letters/view_letter_preview_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+
+    scenario do
```
richrace: We should really add what the scenario is.

```suggestion
    scenario 'Redirect when the user refreshes the Preview Page' do
```

spec/features/letters/view_letter_preview_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+      given_i_am_logged_in
+      when_i_visit_new_letter_page
+      and_i_select letter_type: 'Letter 1 template'
+      and_i_fill_in_the_form_and_submit
+      then_i_see_the_successful_letters_ready_to_send
+      and_i_see_a_send_letter_button
```
richrace: I don't think these are relevant to what this test (scenario) is trying to achieve.

Maybe it should `then_i_am_on_the_preview_page`, linking it to the scenario we're trying to test.

PR #242, opened 2019-11-18
Allow running all services locally

richrace:
### Why

- Allow us to run all services locally

### What 

- Add docker service to allow running from this  docker compose file

### Notes

- run `make run-all` to run all services
- When running all services it's likely you won't see any content in the Worktray. This is due to this line: https://github.com/LBHackney-IT/LBH-IncomeCollection/blob/master/app/controllers/tenancies_controller.rb#L58

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,12 +2,6 @@ INCOME_API_DIR  ?= ../lbh-income-api
 UHSIM_DIR       ?= ../universal-housing-simulator
 TENANCY_API_DIR ?= ../LBHTenancyAPI
 
-ifneq (, $(wildcard ${INCOME_API_DIR}/.env))
--include ${INCOME_API_DIR}/.env
-export
-endif
-
-
 .PHONY: docker-build
 docker-build:
 	docker-compose build
--- a/Makefile
+++ b/Makefile
@@ -39,6 +39,8 @@ check: lint test
 	echo 'Deployable!'
 
 run-all:
+	rm tmp/pids/server.pid || true
+	rm ${INCOME_API_DIR}/tmp/pids/server.pid || true
 	docker build --tag managearrears .
 	cd ${INCOME_API_DIR} && docker build --tag incomeapi .
 	cd ${TENANCY_API_DIR} && docker build --tag tenancyapi -f ./LBHTenancyAPI/Dockerfile .
@@ -53,4 +55,3 @@ shell:
 
 guard:
 	docker-compose run --rm app guard
-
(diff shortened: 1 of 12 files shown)


kjdchapman:
I'm taking a look at this, but it's taking a good while to run locally!

kjdchapman:
If I'm understanding right, this has a dependency on other git repos being on the host machine?

Makefile
```diff
@@ -1,3 +1,13 @@
+INCOME_API_DIR  ?= ../lbh-income-api
+UHSIM_DIR       ?= ../universal-housing-simulator
+TENANCY_API_DIR ?= ../LBHTenancyAPI
+
+ifneq (, $(wildcard ${INCOME_API_DIR}/.env))
```
bjpirt: I did wonder whether we should do this or not - loading in the env variables from the .env file means that you might end up pointing at staging APIs etc, when we can be pretty confident that you're actually trying to run all services locally when running this script (I know I added this initially!)

richrace:
> If I'm understanding right, this has a dependency on other git repos being on the host machine?

Correct

spec/lib/hackney/income/tenancy_gateway_spec.rb
```diff
@@ -282,7 +282,10 @@
               service: "¤#{Faker::Number.decimal(2)}",
               other_charge: "¤#{Faker::Number.decimal(2)}",
               current_arrears_agreement_status: Faker::Lorem.characters(3),
-              current_balance: "¤#{Faker::Number.decimal(2)}",
+              current_balance: {
+                value: "¤#{Faker::Number.decimal(2)}",
+                currency_code: 'GBP'
+              },
```
elena-vi: ermm, has this been changed in the api?
also not comfortable with this change being made here

richrace: This is not a change to the API.

This code was changed: https://github.com/LBHackney-IT/LBH-IncomeCollection/pull/242/files#diff-63b5d8c3f0146e87d16f7d8d02d32203L99
which broke this test.

Makefile
richrace: I can remove loading the `.env` so there's no doubt.

spec/lib/hackney/income/tenancy_gateway_spec.rb
richrace: Had an in-person chat around this.

PR #243, opened 2019-11-18
sending user when sending letters

elena-vi:


elena-vi:
this needs the following backend to work: https://github.com/LBHackney-IT/lbh-income-api/pull/248

PR #246, opened 2019-11-20
Fix sending the wrong formatted params to the API

richrace:
- Using a GET request so should use query rather than form data.
- Have to override `#to_query` as we don't actually serialize the object 
to a Hash first.
- Use a namespace in the `#to_query` to allow us to setup the format of 
the query that the API expects.

### DIFF
--- a/lib/hackney/income/domain/user.rb
+++ b/lib/hackney/income/domain/user.rb
@@ -11,6 +11,10 @@ module Hackney
         def income_collection?
           groups.join(' ').include?('income')
         end
+
+        def to_query(*args)
+          as_json.to_query(*args)
+        end
       end
     end
   end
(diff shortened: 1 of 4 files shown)


lib/hackney/income/domain/user.rb
```diff
@@ -11,6 +11,10 @@ def leasehold_services?
         def income_collection?
           groups.join(' ').include?('income')
         end
+
+        def to_query(*args)
```
maysakanoni: not sure how this works, is this the same as
`def something(user)
 user.as_json 
end`

richrace: The `*args` allows any number of arguments.

Since I'm overriding a method and ultimately calling it again with the same arguments, I do not need to specify the exact arguments in my method signature.

So I can do: `user.to_query`, `user.to_query(:one_arg)`, `user.to_query(1, :something_else, 'or this')` and pass whatever arguements to the real implementation in `as_json.to_query(*args)`.

PR #248, opened 2019-11-20
Add documents view param to documents controller

maysakanoni:
**What?**

- Add param to document controller 

**Why**

- This param is required for the api to determine where the document is being downloaded from. This will determine whether or not the status will be updated to `downloaded` when the documents controller is hit (this is not desired behaviour when the documents controller is hit from the documents view).

### DIFF
--- a/app/views/documents/index.erb
+++ b/app/views/documents/index.erb
@@ -34,7 +34,7 @@
               <td><%= (document[:status] || '').capitalize %></td>
               <td><%= document[:created_at].to_formatted_s(:long_ordinal) %></td>
               <td><%= document[:updated_at].to_formatted_s(:long_ordinal) %></td>
-              <td><%= link_to('Download', document_path(document[:id])) %></td>
+              <td><%= link_to('Download', document_path(id: document[:id], documents_view: true)) %></td>
             </tr>
           <% end %>
         </tbody>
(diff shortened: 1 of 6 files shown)


app/views/documents/index.erb
```diff
@@ -34,7 +34,7 @@
               <td><%= (document[:status] || '').capitalize %></td>
               <td><%= document[:created_at].to_formatted_s(:long_ordinal) %></td>
               <td><%= document[:updated_at].to_formatted_s(:long_ordinal) %></td>
-              <td><%= link_to('Download', document_path(document[:id])) %></td>
+              <td><%= link_to('Download', document_path(id: document[:id], documents_view: true)) %></td>
```
richrace: I don't think you need to specify the `id:` key, the magic Rails routing will assume the first argument is the ID.

spec/controllers/documents_controller_spec.rb
```diff
@@ -20,7 +20,7 @@
 
         expect_any_instance_of(Hackney::Income::DocumentsGateway)
           .to receive(:download_document)
-          .with(id: id, username: @user.name)
+          .with(id: id, username: @user.name, documents_view: nil)
```
richrace: What happens if the param `documents_view` is not supplied to the controller?

richrace: Ignore this. Didn't read the code properly 🤦‍♂

app/views/documents/index.erb
maysakanoni: Yes that works, done here 76632fc

PR #249, opened 2019-11-21
Add param to path in object preview tag to remove duplicate action diary write 

maysakanoni:
**What?**

- Add param to path when documents controller is hit by the object tag 

**Why?**

- There is an object tag in the preview which hits the documents controller. This triggers a premature action diary write. The param is checked in the backend to disable this behaviour.

### DIFF
--- a/app/views/letters/preview.erb
+++ b/app/views/letters/preview.erb
@@ -14,7 +14,7 @@
           <strong>Template name: </strong><%= @preview.dig(:template,:name) %>
           <div class="letter_preview">
             <% if @preview[:document_id] %>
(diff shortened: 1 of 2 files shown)


app/views/letters/preview.erb
```diff
@@ -14,7 +14,7 @@
           <strong>Template name: </strong><%= @preview.dig(:template,:name) %>
           <div class="letter_preview">
             <% if @preview[:document_id] %>
-              <%= tag.object(id: "preview-doc-#{@preview[:document_id]}", data: "#{document_path(@preview[:document_id])}.pdf?inline=true", type: 'application/pdf') %>
+              <%= tag.object(id: "preview-doc-#{@preview[:document_id]}", data: "#{document_path(@preview[:document_id])}.pdf?inline=true&documents_view=true", type: 'application/pdf') %>
```
maysakanoni: is `documents_view` the right param to use?

PR #250, opened 2019-11-22
Add code action diary for LBA

maysakanoni:
**What**

- Add code action diary for LBA

**Why**

- So the action is visible in the front end


richrace:
> It's interesting that this being missing wasn't caught by a test 🤔

There's lots of codes that are not tested - to the backlog?

PR #252, opened 2019-11-25
hide download button

maysakanoni:
**What?**

- Hide download button when a document has been downloaded

**Why?**

- So they can keep track of the documents which have been downloaded as there is no bulk download button.


maysakanoni:
> Talked in person around this JS not being the best. But the work required to not make it ugly is quite a lot.
> 
> For next time, when a file has a lot of style changes (e.g. your editor formatted the code on save) commit that change separately from the new code that you wanted to add. As there's a lot of changed code with a commit message that doesn't really make sense.

I agree, didn't think of that!

PR #255, opened 2019-12-04
display appropriate name for new letter codes

elena-vi:
New letter codes were introduced in UH that we didn't have yet mapped for the front end to display nicely, also I took the opportunity to add a test to make sure that the actions that we've automated display nicely


spec/lib/hackney/income/action_diary_entry_codes_spec.rb
```diff
@@ -51,4 +51,29 @@
       expect(described_class.valid_code?('WON', user_accessible: false)).to eq(true)
     end
   end
+
+  context 'when displaying letter codes' do
```
kjdchapman: I think this test has value, but it won't catch the case we ran into here (adding something to the backend but it doesn't match up to something in the frontend).

Is there anything we could do to prevent it happening again?

PR #261, opened 2019-12-12
Income collection admin

elena-vi:
Our users often need different reports to be generated for their internal use. 

This tool allows the users to export tenancies which have been synced. 

Can be accessed from `income_collection/admin` 
Demo: 
export_gen

Graceful failure demo: 
export_gen_error


kjdchapman:
Can we get an acceptance test around this feature?

Is there a good way to test this javascript?

PR #263, opened 2019-12-12
Add testing around sending a letter

georgeschena:
### WHAT 
Add some testing around sending a letter

### WHY
To give us confidence that what we send to the income-api is what we expect


spec/request/sending_letter_one_manually_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+      expect(response.request.parameters).to eq(
+        'action' => 'send_letter',
+        'controller' => 'income_collection/letters',
+        'tenancy_ref' => tenancy_ref,
+        'uuid' => uuid
+      )
+    end
+
+    it 'will error if no tenancy_ref is provided' do
+      stub_sending_failed_letter_response
```
richrace: We do not need to stub the API as our code does not execute it.

We error on this line: https://github.com/LBHackney-IT/LBH-IncomeCollection/blob/add-testing-around-sending-a-letter/app/controllers/income_collection/letters_controller.rb#L54

georgeschena: 👍

spec/features/income_collection/letters/view_letter_preview_failures_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 describe 'Viewing A Letter Preview' do
   let(:uuid) { SecureRandom.uuid }
-  let(:preview) { Faker::DumbAndDumber.quote }
+  let(:preview) { Faker::TheITCrowd.quote }
```
kjdchapman: It'd be good to justify this change in the commit message for https://github.com/LBHackney-IT/LBH-IncomeCollection/pull/263/commits/9c90283bfda438725ed4538fe698d99ab8fba2b4, I think. We didn't change it for the fun of it!

spec/request/sending_letter_one_manually_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+        'action' => 'send_letter',
+        'controller' => 'income_collection/letters',
+        'tenancy_ref' => tenancy_ref,
+        'uuid' => uuid
+      )
+    end
+
+    it 'will error if no tenancy_ref is provided' do
+      post send_letter_income_collection_letter_path(uuid: uuid, tenancy_ref: nil)
+      expect(flash[:notice]).to eq('Param is missing or the value is empty: tenancy_ref')
```
kjdchapman: Is there a way to check the page content rather than the flash message? 

This smells a little like testing internals, but it might be acceptable in these more "railsy" specs, not sure.

spec/request/sending_letter_one_manually_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  let(:groups) { ['income-collection-group-1'] }
+
+  before do
+    jwt_token = build_jwt_token(groups: groups, user_id: user_id)
+    cookies['hackneyToken'] = jwt_token
+  end
+
+  context 'with a income collection user' do
+    it 'will send a letter successfully if a uuid and tenancy_ref is provided' do
+      stub_sending_successful_letter_response
```
kjdchapman: I'd tend to pull out the collaborator stubbing to sit next to a context, so that this test shouts about the mocking its doing, e.g.:

```ruby
describe 'sending a letter manually', do
  context 'with a stubbed income api letter endpoint' do
    before { stub_request(:post, '/blah', ).to_return(result)
    
    context 'that returns success' do
      let(:result) { :success }
      
      it 'shows the send was successful' do
        expect(page).to have_content('yay!')
      end
    end

    context 'that returns an error' do
      let(:result) { :internal_server_error }
      
      it 'shows the send failed' do
        expect(page).to have_content('oh noes!')
      end
    end
  end
end
```

spec/request/sending_letter_one_manually_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+    it 'will error if no tenancy_ref is provided' do
+      post send_letter_income_collection_letter_path(uuid: uuid, tenancy_ref: nil)
+      expect(flash[:notice]).to eq('Param is missing or the value is empty: tenancy_ref')
+    end
+  end
+
+  private
+
+  def stub_sending_successful_letter_response
+    stub_request(:post, 'https://example.com/income/apiv1/messages/letters/send')
```
kjdchapman: This path looks incorrect (no space between api and v1) - typo?

richrace: This is a fun one. It's actually "correct". This is due to us manually building the URI instead of using the `URI` class and building it in a consistent way.

This line: 

https://github.com/LBHackney-IT/LBH-IncomeCollection/blob/master/.env.test#L2

and this line: 

https://github.com/LBHackney-IT/LBH-IncomeCollection/blob/master/lib/hackney/income/letters_gateway.rb#L16

Mean we do not check to see where the `/` should go. In the Env variables for staging we put the API URI as `api/`

PR #264, opened 2019-12-16
Timeline improvements

elena-vi:
What's been done: 
- [x] Make a timeline header stay in place when a user scrolls through the timeline.
- [x] Don’t show all outgoing transactions
- [x] Add balance to the diary action entries
- [x] show the part of the action diary detail e.g. 1st line.
- [x] Only have two checkboxes for filtering; show all action diary entries and show all transactions rather than the current 4 options we have.

DEMO:
Dec-16-2019 15-33-18


app/helpers/tenancy_helper.rb
```diff
@@ -20,4 +20,35 @@ def pause_reasons
...
+        <details class="govuk-details" data-module="govuk-details">
+          <summary class="govuk-details__summary">
+            <span class="govuk-details__summary-text">
+              Continue reading
+            </span>
+          </summary>
+          <div class="govuk-details__text">#{comment}</div>
+        </details>
+        HTML
+      .html_safe
```
richrace: Can we extract this into a partial ERB and use that instead?

app/views/common/_tenancy_timeline_list.html.erb
```diff
@@ -70,14 +70,17 @@
           <tr class="transactions-table--action transactions-table__group--breakdown" data-balance="<%= item[:balance] %>" data-action="<%= item[:action_diary_balance] %>">
             <td><%= format_short_date(item[:date]) %></td>
             <td><%= Hackney::Income::ActionDiaryEntryCodes.human_readable_action_code(item[:code]) %></td>
-            <td colspan="2"><%= item[:comment]%></td>
+            <td colspan="2">
+              <%= render partial: 'action_diary_entry', locals: { action_diary_comment: item[:comment] } %>
```
richrace: The partial name and the variable it takes do not match to me.

app/views/common/_tenancy_timeline_list.html.erb
```diff
@@ -70,14 +70,17 @@
...
+            </td>
             <td><%= item[:user]%></td>
             <td>
+              <%= number_to_currency(item[:action_diary_balance], unit: '£') %>
               <div class="type-label"><strong>ACTION</strong></div>
             </td>
           </tr>
-        <% elsif item[:transaction] %>
-          <tr class="transactions-table--transaction transactions-table__group--breakdown" data-balance="<%= item[:balance] %>">
+        <% elsif item[:transaction] && ( item[:type] == 'DBR' || item[:value].negative? )%>
```
richrace: Can we extract `item[:type] == 'DBR' || item[:value].negative?` into a method. I don't understand the importance of these values.

Is it `transaction_is_payment_or_x?` ?

app/views/common/_tenancy_timeline_list.html.erb
```diff
@@ -92,6 +95,38 @@
...
+          <details class="govuk-details transactions-table__expand-link show_table" data-expand="hidden_transactions" data-period="<%= group_summary[:date_range] %>">
+            <summary class="govuk-details__summary">
+              <span class="govuk-details__summary-text">
+                Other charges
+              </span>
+            </summary>
+          </details>
+        </td>
+        <td class="numeric <%= class_for_value(1) %>">
+          <%= number_to_currency((group_summary[:list].select {|i| i[:transaction] && i[:type] != 'DBR' && i[:value].positive? }.sum { |i| i[:value] }), unit: '£')%>
```
richrace: What's the magic value of `DBR`? I think this sum would be more readable as a function.

app/views/common/_tenancy_timeline_list.html.erb
```diff
@@ -92,6 +95,38 @@
...
+        </td>
+        <td></td>
+        <td></td>
+      </tr>
+      <% group_summary[:list].select {|i| i[:transaction] && i[:type] != 'DBR' && i[:value].positive? }.each do |item| %>
```
richrace: Similar theme here.

PR #270, opened 2020-01-06
Make sure the User's JSON string is consistent

richrace:
When we test API calls we send a JSON User object. Sometimes the order 
of the fields is inconsistent. This breaks our tests. Always ensure we 
have the consistent JSON string. See the following Job for an example of 
a random failing test

https://app.circleci.com/jobs/github/LBHackney-IT/LBH-IncomeCollection/1576

### DIFF
--- a/spec/lib/hackney/income/domain/user_spec.rb
+++ b/spec/lib/hackney/income/domain/user_spec.rb
@@ -187,7 +187,7 @@ describe Hackney::Income::Domain::User do
       "{\"id\":\"#{id}\",\"name\":\"#{name}\",\"email\":\"#{email}\",\"groups\":[#{formatted_groups}]}"
     end
 
-    it 'returns a consistent json string' do
+    it 'returns a consistent json string regardless of rspec seed' do
       expect(user.to_json).to eq(expected_json_string)
     end
   end
--- a/spec/lib/hackney/income/domain/user_spec.rb
+++ b/spec/lib/hackney/income/domain/user_spec.rb
@@ -141,4 +141,54 @@ describe Hackney::Income::Domain::User do
       end
     end
   end
+
+  describe '#as_json' do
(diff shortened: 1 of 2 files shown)


kjdchapman:
If I understand correctly, this test is a "canary" - that is to say, it'll fail if the random seed is one that causes the weird JSON ordering behaviour we talked about earlier this week.

Have you found seed values where both this test and the other tests fail?

richrace:
Two seeds I'm aware of are:
- 9661
- 31813

They only fail this test.

kjdchapman:
If only this test fails, I don't think I'm understanding things correctly!

What's the benefit of having this test?

richrace:
Using the seeds if you move the following line to be after the `.new.tap ...`, the failing test passes but others fail. 

https://github.com/LBHackney-IT/LBH-IncomeCollection/blob/master/spec/lib/hackney/income/domain/user_spec.rb#L100

spec/lib/hackney/income/domain/user_spec.rb
```diff
@@ -141,4 +141,54 @@
...
+        user.id = id
+      end
+    end
+    let(:expected_json_string) do
+      formatted_groups = groups.map { |g| "\"#{g}\"" }.join(',')
+
+      "{\"id\":\"#{id}\",\"name\":\"#{name}\",\"email\":\"#{email}\",\"groups\":[#{formatted_groups}]}"
+    end
+
+    it 'returns a consistent json string' do
```
richrace:
```suggestion
    it 'returns a consistent json string regardless of rspec seed' do
```

kjdchapman:
We talked about this in-person - I'd got completely the wrong idea. 

This isn't a canary, it's a new implementation of the `to_json` method, as the default implementation behaves inconsistently between test runs.

PR #279, opened 2020-01-17
Add Paused Reason "Missing Data"

richrace:
### Why

There's some bad data in UH.  So we want to pause cases whilst the correct data is updated. So it does not fall into the automation process.

### What

Add a new Reason and Action Diary Code


richrace:
I've tested it on Local Front  - Staging Back.

PR #302, opened 2020-06-02
Update ruby from Ruby 2.5.3 to Ruby 2.7.1

C-gyorfi:
## Context
We are using Ruby 2.5.3 was released in October 2018 - we should try updating Ruby and see if it causes any issues

## Changes proposed in this pull request

- Update Ruby from `2.5.3` to `2.7.1`
- Update rubocop -> the older version is not compatible with ruby 2.7.X
- Resolve linting after the rubocop update

## Guidance to review
Although the tests are passing and I had no issues running the app locally, it would be great if you could pull this branch and try too! 

There are a couple of warnings related to `warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call` 
These should be fixed by updating the rails version and other related gems, I'll do them in a separate PR if that's ok.

## Link to Jira card

https://hackney.atlassian.net/secure/RapidBoard.jspa?rapidView=30&projectKey=MAAP&modal=detail&selectedIssue=MAAP-164


bjpirt:
Ideally we'd try the 2.7 branch of Ruby since that is the latest (2.7.1), but if we've decided to go with 2.6 branch for some reason, we should probably use the latest sub-version on that (2.6.6)

Did you try 2.7.1?

C-gyorfi:
> Ideally we'd try the 2.7 branch of Ruby since that is the latest (2.7.1), but if we've decided to go with 2.6 branch for some reason, we should probably use the latest sub-version on that (2.6.6) - 2.6.3 was release over a year ago
> 
> Did you try 2.7.1?

Thanks, I'll try to update to 2.7.1!

PR #304, opened 2020-06-04
display an error message to the user when manual sms messages fails

elena-vi:
## Context

The user doesn't know when their sms failed to send

## Changes proposed in this pull request

Catch errors coming out from the Income API and display them in a flash message

## Link to Jira card

https://hackney.atlassian.net/browse/MAAP-206

## Things to check

- [ ] If this code includes a migration adding or changing columns, it also backfills existing records for consistency


app/controllers/tenancies_sms_controller.rb
```diff
@@ -12,14 +12,18 @@ def create
...
+      use_cases.send_sms.execute(
+        phone_numbers: phone_numbers,
+        tenancy_ref: params.fetch(:id),
+        template_id: params.fetch(:template_id),
+        username: current_user.name
+      )
+      flash[:notice] = 'Successfully sent the tenant an SMS message'
+      redirect_to tenancy_path(id: params.fetch(:id))
+    rescue Exceptions::IncomeApiError::UnprocessableEntity => e
+      flash[:notice] = e.message
```
C-gyorfi, line 25: Just wondering if this is a "user friendly" message?

PR #308, opened 2020-06-17
Add create agreement page

C-gyorfi:
## Context
We want to allow caseworkers to add an agreement so that tenants can repay their arrears over an agreed term. This is the first slice (or second if we count the feature flag system https://github.com/LBHackney-IT/LBH-IncomeCollection/pull/311), to build the UI for agreements. Next tasks:
- Add JS calculator to the create agreement page *see guidance to review
- Add some sort of input validation to the form *see guidance to review
- Agreement details page
- Agreements history page 

- Add `create agreement` button to `case profile` page
- Implement `create agreement` page
- Show new agreement on the case profile page

### After
image

image

image

## Guidance to review
There are a couple of things missing from the page but given the size of the PR, I think we should iterate on this. Some things that are missing:
- The test only covers the happy path(e.g we should also test for cases when the user enters invalid values)
- We need to think about what validations we want to use, for instance, what happens if the instalment amount is too low or greater than the balance? Is there a min or max limit?
- The `end date` calculator is missing, this will be some JS magic, although I added some JS for the field label, the calculator is a bit more complicated and should be unit tested and reusable. Let's tackle that in a separate PR.
- The notes are not being recorded, because it is unclear at the moment where and how we want to record these.

### Guidance for manual testing
- Visit the feature-flag path and switch on the feature
- Visit a case profile
- Click on the create an agreement
- And ask me if you need any more guidance

## Link to Jira card
https://hackney.atlassian.net/secure/RapidBoard.jspa?rapidView=30&projectKey=MAAP&modal=detail&selectedIssue=MAAP-173

## Things to check

### DIFF
--- a/app/controllers/agreements_controller.rb
+++ b/app/controllers/agreements_controller.rb
@@ -1,5 +1,6 @@
 class AgreementsController < ApplicationController
   def new
     @tenancy_ref = params.fetch(:tenancy_ref)
+    @tenancy = use_cases.view_tenancy.execute(tenancy_ref: @tenancy_ref)
   end
 end
--- a/app/controllers/agreements_controller.rb
+++ b/app/controllers/agreements_controller.rb
@@ -1,4 +1,6 @@
 class AgreementsController < ApplicationController
+  before_action { redirect_to worktray_path unless FeatureFlag.active?('create_informal_agreements') }
+
   def new
     @tenancy_ref = params.fetch(:tenancy_ref)
     @tenancy = use_cases.view_tenancy.execute(tenancy_ref: @tenancy_ref)
--- a/app/controllers/agreements_controller.rb
+++ b/app/controllers/agreements_controller.rb
@@ -1,0 +1,5 @@
+class AgreementsController < ApplicationController
+  def new
+    @tenancy_ref = params.fetch(:tenancy_ref)
+  end
+end
--- a/app/controllers/agreements_controller.rb
+++ b/app/controllers/agreements_controller.rb
@@ -3,4 +3,17 @@ class AgreementsController < ApplicationController
     @tenancy_ref = params.fetch(:tenancy_ref)
     @tenancy = use_cases.view_tenancy.execute(tenancy_ref: @tenancy_ref)
   end
+
+  def create
+    tenancy_ref = params.fetch(:tenancy_ref)
+    use_cases.create_agreement.execute(
+      tenancy_ref: tenancy_ref,
+      frequency: params.fetch(:frequency).downcase,
+      amount: params.fetch(:instalment_amount),
+      start_date: params.fetch(:start_date)
+    )
+
+    flash[:notice] = 'Successfully created a new agreement'
+    redirect_to tenancy_path(id: tenancy_ref)
+  end
 end
--- a/app/controllers/tenancies_controller.rb
+++ b/app/controllers/tenancies_controller.rb
@@ -41,6 +41,11 @@ class TenanciesController < ApplicationController
     tenancy_ref = params.fetch(:id)
     @tenancy = use_cases.view_tenancy.execute(tenancy_ref: tenancy_ref)
 
+    if FeatureFlag.active?('create_informal_agreements')
+      agreements = use_cases.view_agreements.execute(tenancy_ref: tenancy_ref)
+      @agreement = agreements.find { |agreement| agreement.current_state == 'live' }
+    end
+
     render :show
   end
 
--- a/app/views/agreements/new.html.erb
+++ b/app/views/agreements/new.html.erb
@@ -25,7 +25,7 @@
       </div>
 
       <div class="form-group">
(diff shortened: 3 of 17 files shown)


app/controllers/agreements_controller.rb
```diff
@@ -0,0 +1,21 @@
+class AgreementsController < ApplicationController
+  before_action { redirect_to worktray_path unless FeatureFlag.active?('create_informal_agreements') }
+
+  def new
+    @tenancy_ref = params.fetch(:tenancy_ref)
+    @tenancy = use_cases.view_tenancy.execute(tenancy_ref: @tenancy_ref)
```
antonyoneill, line 6: So, just as I'm not overly familiar with Rails.

This will automatically render the AgreementsController new view, and `@tenancy_ref` and `@tenancy` will be available for the view to render properties?

In essence this controller is passing a domain object into the html template, and that maps the domain object to a presentation view

app/controllers/tenancies_controller.rb
```diff
@@ -41,6 +41,11 @@ def show
     tenancy_ref = params.fetch(:id)
     @tenancy = use_cases.view_tenancy.execute(tenancy_ref: tenancy_ref)
 
+    if FeatureFlag.active?('create_informal_agreements')
+      agreements = use_cases.view_agreements.execute(tenancy_ref: tenancy_ref)
+      @agreement = agreements.first { |agreement| agreement.current_state == 'live' }
```
antonyoneill, line 46: Nice, can you talk me through this syntax? I think I'm missing something :grimacing: [Array.first](https://ruby-doc.org/core-2.7.1/Array.html#method-i-first) doesn't take a block like this, is it a ruby thing?

app/views/agreements/new.html.erb
```diff
@@ -0,0 +1,63 @@
...
+      </div>
+
+      <%= submit_tag 'Create', class: 'button' %>
+    <% end %>
+  </div>
+</div>
+
+
+<script type="text/javascript">
+  document.getElementById("start_date").addEventListener('change', (event) => {
```
antonyoneill, line 54: :see_no_evil: I know this is the rails way of doing things - but seeing raw javascript like this feels bad to me (coming from a stack with a proper JS build pipeline) - not least because it's not unit tested, but also it is not reusable.

Anyway, in this particular example I'd be keen for us to avoid using `innerHTML` as it does carry some [security vulnerabilities](https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML#Security_considerations). I'd recommend changing to be `textContent` instead. It would also be good to remove the HTML tags from the content that you set, so consider adding those HTML tags to the 'original' and using a `<span/>` element to put the content in.

lib/hackney/income/create_agreement_gateway.rb
```diff
@@ -0,0 +1,33 @@
+require 'uri'
+require 'net/http'
+
+module Hackney
+  module Income
+    class CreateAgreementGateway
```
antonyoneill, line 6: I like that you have a separate gateway for this! I've usually seen singular API endpoint gateway which has several functions on it that operate on different objects

lib/hackney/income/view_agreements_gateway.rb
```diff
@@ -0,0 +1,38 @@
+require 'uri'
+require 'net/http'
+
+module Hackney
+  module Income
+    class ViewAgreementsGateway
+      def initialize(api_host:, api_key:)
+        @api_host = api_host
```
antonyoneill, line 8: I wonder if some of this can be abstracted up into a general API gateway module with individual classes?

Or if the AgreementsGateway should be combined to operate on `view_agreements` `create_agreement` - separating out different aspects of the same API makes sense, but I wonder if separating out gateways for the same aspect is too far? Food for thought

app/controllers/agreements_controller.rb
C-gyorfi: Well described 👌  but we can have a chat about this too!

app/controllers/tenancies_controller.rb
C-gyorfi: Good catch what I wanted to do is
```Ruby
agreements.find { |agreement| agreement.current_state == 'live' }
```
it passed the test for the wrong reason

app/views/agreements/new.html.erb
C-gyorfi: I don't know if this is the rails way, honestly, this is the first time I'm writing JS in an erb file and clearly this was is a hacky solution so thanks for the suggestion! I changed this in ccfb0c3

Also, I think we could improve the way we are building UI. I've added a [ticket to install node modules for the GOV.UK Frontend gem](https://hackney.atlassian.net/browse/MAAP-249). There is a `govuk_frontend_toolkit` gem installed that is already depreciated and does not meet the gov accessibility standards. The CSS and JS do not render for the HTML classes and therefore many of the UI components are not using it the `govuk` components. For instance the date fields we are using. We could aim to build UI that's more aligned with GOV UK design system.

PR #309, opened 2020-06-18
Add make command for syncing income api data

Ninamma:
## Context

Currently when you want to run the whole application with `make run-all`, you'll have to go into the incomeapi container and run commands to populate the incomeapi with data.

I've popped those two commands into the Makefile so after `make run-all` you can just run `make sync` to get all the data.

## Guidance to review

This could live in pretty much any of the repos since it'll just pick up the container name - I think this repo or incomeapi repo make the most sense.

I thought to put it here because I tend to need it most after running `make run-all`.

But since it mainly concerns incomeapi, I can see that it might make sense to be in that repo.

Or both? 🤔 

Also I'm not too sure if I've completely understood what the commands do - so if there's a better name than `sync`, please let me know!
## Things to check

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,5 +52,5 @@ shell:
 guard:
 	docker-compose run --rm app guard
 
-sync:
+sync-uh-simulator-data:
 	docker exec -ti universal-housing-simulator_incomeapi_1 sh -c "export CAN_AUTOMATE_LETTERS=true && rake income:sync:manual_sync"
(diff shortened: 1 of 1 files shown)


C-gyorfi:
> Also I'm not too sure if I've completely understood what the commands do - so if there's a better name than `sync`, please let me know!

That is right, `sync` is very generic maybe `sync-uh-simulator-data` ?

PR #311, opened 2020-06-19
Add a feature flag system that uses env vars

C-gyorfi:
## Context

We want to be able delivering thin slices of a larger feature and switch
on the feature when it is ready to rollout. We also want to be able to keep track of features that are not in production yet.

I was going to use [rollout gem](https://github.com/fetlife/rollout) that has a lot of capabilities for implementing a feature flag system using redis, but there was some issue with the communication with Redis server from front-end so I ended up using ENV variables instead, should be enough for what we trying to achieve.

- Add a feature flag system that allows as to define a feature flag and switch it on/off
- Add `create_informal_agreement' flag
- Add a dashboard that allows non-developers to switch on a feature e.g in staging and test them before we release them into prod.

image

## Guidance to review

Are we happy with this approach? 

## Link to Jira card
https://hackney.atlassian.net/secure/RapidBoard.jspa?rapidView=30&projectKey=MAAP&modal=detail&selectedIssue=MAAP-248
## Things to check

### DIFF
--- a/app/controllers/feature_flags_controller.rb
+++ b/app/controllers/feature_flags_controller.rb
@@ -1,0 +1,13 @@
+class FeatureFlagsController < ApplicationController
+  def index; end
+
+  def activate
+    FeatureFlag.activate(params[:feature_name])
+    redirect_to feature_flags_dashboard_path
+  end
+
+  def deactivate
+    FeatureFlag.deactivate(params[:feature_name])
+    redirect_to feature_flags_dashboard_path
+  end
+end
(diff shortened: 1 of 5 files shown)


C-gyorfi:
> * Is it considered safe in rails to be mutating the ENV? Remember that it would only change those variables for this instance of the rails server
> * What happens if we redeploy/the ECS task is cleaned and reinitialised? Would the user preferences be lost?

Not sure how safe it is, I can have a look on Monday, as far as I know, the front-end is deployed on Heroku. 

> * Who should be allowed access to the feature flags view? (We may not care about this)

Currently, anyone can access the dashboard who is authorised to use the service, maybe we can add some restrictions if it's necessary

PR #314, opened 2020-06-26
Allow caseworkers to view agreement details

C-gyorfi:
## Context
We want to allow caseworkers to add an agreement so that tenants can repay their arrears over an agreed term. This is the third slice, to build the UI for agreements -> in which we add the `Agreement details` page.
Next tasks:
- Add cancel button(this also needs a new API endpoint)
- Agreements history page 
- Implement notes 
- Add JS calculator to the create agreement page 

- Add 'view details` link to agreements partial
- Render `agreement details` page
- Show `Created by` and `Created at` information, `agreement details` and `agreement state history`

## Link to Jira card
https://hackney.atlassian.net/secure/RapidBoard.jspa?rapidView=30&projectKey=MAAP&modal=detail&selectedIssue=MAAP-173

## Things to check

### DIFF
--- a/app/controllers/agreements_controller.rb
+++ b/app/controllers/agreements_controller.rb
@@ -2,12 +2,10 @@ class AgreementsController < ApplicationController
   before_action { redirect_to worktray_path unless FeatureFlag.active?('create_informal_agreements') }
 
   def new
-    @tenancy_ref = params.fetch(:tenancy_ref)
-    @tenancy = use_cases.view_tenancy.execute(tenancy_ref: @tenancy_ref)
+    @tenancy = use_cases.view_tenancy.execute(tenancy_ref: tenancy_ref)
   end
 
   def create
-    tenancy_ref = params.fetch(:tenancy_ref)
     use_cases.create_agreement.execute(
       tenancy_ref: tenancy_ref,
       frequency: params.fetch(:frequency).downcase,
@@ -20,6 +18,18 @@ class AgreementsController < ApplicationController
   end
 
   def show
-    @tenancy_ref = params.fetch(:tenancy_ref)
+    @tenancy = use_cases.view_tenancy.execute(tenancy_ref: tenancy_ref)
+    @agreement = use_cases.view_agreements.execute(tenancy_ref: tenancy_ref)
+                .find { |agreement| agreement.id == agreement_id }
+  end
+
+  private
+
+  def tenancy_ref
+    @tenancy_ref ||= params.fetch(:tenancy_ref)
+  end
+
+  def agreement_id
+    @agreement_id ||= params.fetch(:id).to_i
   end
 end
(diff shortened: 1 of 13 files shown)


app/controllers/agreements_controller.rb
```diff
@@ -18,4 +16,20 @@ def create
...
+  def show
+    @tenancy = use_cases.view_tenancy.execute(tenancy_ref: tenancy_ref)
+    @agreement = use_cases.view_agreements.execute(tenancy_ref: tenancy_ref)
+                .find { |agreement| agreement.id == agreement_id }
+  end
+
+  private
+
+  def tenancy_ref
+    @tenancy_ref ||= params.fetch(:tenancy_ref)
```
antonyoneill, line 29: Nice!

app/views/agreements/_agreement_status_history.html.erb
```diff
@@ -0,0 +1,17 @@
+<table class="agreements_history-table fixed-layout">
+  <thead>
+    <th class="date-column">Date</th>
+    <th class="status-column">Status</th>
+    <th class="descreption-column">Descreption</th>
+  </thead>
+  <% agreement_status_history.each do |state| %>
```
antonyoneill, line 7: Could we call the parameter `status_history` or something rather than `state`?

app/views/tenancies/case/_agreements.html.erb
```diff
@@ -7,11 +7,22 @@
             <h2>Arrears Agreement</h2>
             There are currently no live agreement
           <% else %>
-            <h2>Arrears Agreement</h2>
-            <label class="govuk-label" for="status"><strong>Status: </strong><%= @agreement.current_state.humanize %><br/></label>
-            <label class="govuk-label" for="status"><strong>Expected balance: </strong><%= number_to_currency(@agreement.starting_balance, unit: '£') %><br/></label>
-            <label class="govuk-label" for="status"><strong>Actual balance: </strong><%= number_to_currency(@agreement.starting_balance, unit: '£') %><br/></label>
-            <label class="govuk-label" for="status"><strong>Last checked: </strong><br/></label>
+          <div class="grid-row">
```
antonyoneill, line 10: Does rubocop run against erb files? Just a very minor indentation issue here

spec/features/income_collection/agreements/creating_informal_agreement_spec.rb
```diff
@@ -135,13 +186,19 @@ def stub_view_agreements_response
             "agreementType": 'informal',
             "startingBalance": '103.57',
             "amount": '50',
-            "startDate": '2020-06-19',
+            "startDate": '2020-12-12',
```
antonyoneill, line 189: Do these need to be dynamic dates? Would tests start failing after December?

C-gyorfi: Not with the current implementation, because it's just a dummy data stubbing out the API response, but I think it could cause issues when we have end date calculator implemented, not sure if we are going to test that in this spec though. Let's cross this bridge later!

app/views/tenancies/case/_agreements.html.erb
C-gyorfi: Nope, there is an alternative called ERB lint, but it is super slow and the pain outweighs the added value 😂

PR #317, opened 2020-07-01
Allow caseworkers to view agreements history

Ninamma:
## Context

We want caseworkers to be able to view the history of agreements for a tenant.
This can help them better understand the tenant's situation.

- Add and render`History of agreements` page 
  - Description column has latest state of an agreement and date when that state was created - unless it's 'live'
image

- Allow users to see history of agreements even if there are no 'live' agreements
  - Previously the `View history` link was hidden if there weren't any 'live' agreements
image

- Updated the Makefile's sync command since we've [updated the namesapce in the income-api](https://github.com/LBHackney-IT/lbh-income-api/pull/421)
image

Things to do:
- Implement `End date` calculation
- The design has all dates in DD/MM/YYYY format, however I've used an existing method to keep dates as e.g. `July 7th, 2020` 
  - So if we need to match the design, we'll need to add a new date format method in `application_helper`
  - But I wasn't sure if the existing date formats are the ones preferred by users 🤔 

## Guidance to review

I don't have much experience with "convention" so would love 👀 on things like the way I've named variables/files/methods.

I've also moved around the logic in the erb for the _agreements partial, so a check on that would be great!
## Link to Jira card

https://hackney.atlassian.net/browse/MAAP-252?atlOrigin=eyJpIjoiYjE3NzdmNjdhZTIyNDg2MTk4YzIzOGNlYTJkMDc2NjAiLCJwIjoiaiJ9
## Things to check


app/controllers/tenancies_controller.rb
```diff
@@ -42,8 +42,8 @@ def show
     @tenancy = use_cases.view_tenancy.execute(tenancy_ref: tenancy_ref)
 
     if FeatureFlag.active?('create_informal_agreements')
-      agreements = use_cases.view_agreements.execute(tenancy_ref: tenancy_ref)
-      @agreement = agreements.find { |agreement| agreement.current_state == 'live' }
+      @agreements = use_cases.view_agreements.execute(tenancy_ref: tenancy_ref)
```
C-gyorfi, line 45: I'm a bit confused about this change?

Ninamma: In `app/views/tenancies/case/_agreements.html.erb:18` I wanted to be able to show the 'View History' link when there are previous agreements but no live agreeements.

Currently you won't see that link in that situation.

So to check whether there are any agreements, I changed `agreements` to `@agreements` - which I think means I can access it and check it in the erb? And added a check for it in the _agreements partial

app/views/tenancies/case/_agreements.html.erb
```diff
@@ -3,25 +3,28 @@
...
                   <%= link_to('View details', show_agreement_path(tenancy_ref: @tenancy.ref, id: @agreement.id)) %>
                 </div>
+              <% end %>
+              <% if @agreements.present? %>
                 <div class="column-align-right">
-                  <%= link_to('View history') %>
+                  <%= link_to('View history', show_agreements_history_path(tenancy_ref: @tenancy.ref)) %>
                 </div>
-              </div>
+              <% end %>
```
C-gyorfi, line 22:
```RUBY
<% if @agreement.present? %>
   <div class="column-half">
         <%= link_to('View details', show_agreement_path(tenancy_ref: @tenancy.ref, id: @agreement.id)) %>
    </div>
    <div class="column-align-right">
        <%= link_to('View history') %>
        <%= link_to('View history', show_agreements_history_path(tenancy_ref: @tenancy.ref)) %>
    </div>
 <% end %>
```

We could cut down one of the if statements, although I think we want to show the agreement history link even when there is no active agreement 🤔

Ninamma: yup so before we only had one if statement for `@agreement.present?`, but I added another check for `@agreements.present` so that the history link could still be seen

C-gyorfi: Oh I think I just understood why did you change the tenancy controller 👍  ignore me 😃

C-gyorfi: Could do if, elsif, else but its not a big deal :)

spec/features/income_collection/agreements/creating_informal_agreement_spec.rb
```diff
@@ -35,13 +35,15 @@
...
     and_i_should_see_a_link_to_view_details
+    and_i_should_see_a_link_to_view_history
 
     when_i_click_on_view_details
     then_i_should_see_the_agreement_details_page
     and_i_should_see_the_agreement_status
     and_i_should_see_the_agreement_details
     and_i_should_see_a_button_to_cancel_and_create_new_agreement
     and_i_should_see_the_agreement_state_history
+
```
C-gyorfi, line 46: Was this extra line intentional?

spec/features/income_collection/agreements/creating_informal_agreement_spec.rb
```diff
@@ -149,7 +162,36 @@ def when_i_confirm_to_cancel_the_agreement
   end
 
   def and_i_should_not_see_a_live_agreement
-    expect(page).to have_content('There are currently no live agreement')
+    expect(page).to have_content('There are currently no live agreements')
```
C-gyorfi, line 165: I think in the future only one agreement can be live at a time, but we could check with Jenny what is the best content for this

spec/features/income_collection/agreements/creating_informal_agreement_spec.rb
```diff
@@ -235,14 +277,46 @@ def stub_view_agreements_response
...
       .with(
         headers: { 'X-Api-Key' => ENV['INCOME_API_KEY'] }
       )
       .to_return({ status: 200, body: response_with_no_agreements_json },
-                 { status: 200, body: response_with_agreement_json },
-                 { status: 200, body: response_with_agreement_json },
-                 status: 200, body: response_with_no_agreements_json)
+                 { status: 200, body: response_with_live_agreement_json },
+                 { status: 200, body: response_with_live_agreement_json },
+                 status: 200, body: response_with_cancelled_agreement_json)
```
C-gyorfi, line 319: Thanks for fixing this 👍

app/views/tenancies/case/_agreements.html.erb
Ninamma: ah okay - I don't quite see how if, elsif and else would work in the situation 🤔  but yeah if it's not a big deal we can look at it later 😄

spec/features/income_collection/agreements/creating_informal_agreement_spec.rb
Ninamma: Ahh okay, that makes sense, because of the 'there are' I was a bit confused, so I will change it back, sorry!

app/controllers/tenancies_controller.rb
```diff
@@ -42,8 +42,8 @@ def show
...
+      @agreement = @agreements.find { |agreement| agreement.current_state == 'live' }
```
antonyoneill, line 46: Sorry to wade in on an already merged PR - it would be good to rename this at a later date to `@live_agreement` owtta so it's clear later in the view.

(No change required)

PR #319, opened 2020-07-03
Rescue from ActionController::InvalidAuthenticityToken

C-gyorfi:
## Context
image

This occurs when calling the create action in agreements controller, potentially caused by the page caching. "If you use page caching, Rails will keep the static HTML for a page in the cache. After a while the authenticity token for this page is stale and posting a form will result in this error. In this case either don't use page caching, or disable the forgery protection for that controller action" [Source](https://appsignal.com/for/invalid_authenticity_token)

- Add `protect_from_forgery` to AgreementsController 

## Guidance to review 
Adding `protect_from_forgery` to AgreementsController fixed this issue, but the success flash message is no longer shown because this flash message depends on the session cookies

Alternatively we could use:

`skip_before_action :verify_authenticity_token, :only => [:create]`

As far as I understand, this would introduce some security issues as it would disable CSRF (Cross Site Request Forgery) protection for this request. Although the API only accepts calls from the Fron-end application, this may be an issue in the future if we'd allow other requests. 

What do you think?

## Things to check

### DIFF
--- a/app/controllers/agreements_controller.rb
+++ b/app/controllers/agreements_controller.rb
@@ -4,6 +4,7 @@ class AgreementsController < ApplicationController
 
   def new
     @tenancy = use_cases.view_tenancy.execute(tenancy_ref: tenancy_ref)
+    @start_date = (Date.today + 1.day).to_s if @start_date.nil?
   end
 
   def create
--- a/app/controllers/agreements_controller.rb
+++ b/app/controllers/agreements_controller.rb
@@ -15,7 +15,13 @@ class AgreementsController < ApplicationController
       start_date: params.fetch(:start_date),
       created_by: @current_user.name
     )
+    redirect_to show_success_path
+  rescue Exceptions::IncomeApiError => e
+    flash[:notice] = "An error occurred: #{e.message}"
+    render :new, tenancy_ref: tenancy_ref
+  end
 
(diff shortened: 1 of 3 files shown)


C-gyorfi:
> I'm not sure why the flash would stop working as a result of this though - maybe we should investigate that a bit further

Based on this [issue](https://github.com/rails/rails/issues/20035) it's the expected behaviour, but I'll have a look at how to workaround 💭

antonyoneill:
I think I must be missing something, so reading between the lines the token is set in a cookie (in previous roles it's set on the page and passed back as a `hidden` input field).

But an invalid token doesn't necessarily mean the cookies aren't working - it could just be that there's a malicious actor trying to perform a CSRF attack, or that the person has had the form open for too long and the token has expired, or a developer has forgotten to initialise the token in the first place :wink: 

Does rails usually allow flash messages on a redirect? If yes, it seems really odd that in this particular example that it would ignore the flash cookie.

C-gyorfi:
> Does rails usually allow flash messages on a redirect? If yes, it seems really odd that in this particular example that it would ignore the flash cookie.

Yes, I don't think I fully understand the problem, but it allows to flash a success message from a new controller action 2925df7 not sure how I feel about this fix...

PR #321, opened 2020-07-07
Record agreement notes when creating new agreements

C-gyorfi:
## Context

We want to allow caseworkers to add notes when creating a new agreement

- Start recording notes on create agreement form
- Render notes when reviewing an existing agreement
- Refactor agreements controller to persist form data when the create
agreement request blows up, so users don't need to refill the form to
resubmit

### After
image

## Link to Jira card

https://hackney.atlassian.net/browse/MAAP-267

## Things to check

### DIFF
--- a/app/controllers/agreements_controller.rb
+++ b/app/controllers/agreements_controller.rb
@@ -4,21 +4,22 @@ class AgreementsController < ApplicationController
 
   def new
     @tenancy = use_cases.view_tenancy.execute(tenancy_ref: tenancy_ref)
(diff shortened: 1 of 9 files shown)


C-gyorfi:
> * The page refreshed but the Create agreement form looked exactly the same, without any messages

Oh, I think its because of the `protect_from_forgery`

> But when I went back to the case profile, the agreement was there and the notes were all fine!

Yes I think because it crashed when creating the action diary entry, its not ideal, but I tested on staging and worked

PR #327, opened 2020-07-20
adding skip to main content link

LBHLMatthewsMurray:
## Context
Adding skip to main content link to help keyboard-only users skip to the main content on a page.

- Adding skip to main content link component to each page on Manage arrears
- Updating application.html.erb

Screenshot 2020-07-28 at 12 30 39

## Guidance to review
https://design-system.service.gov.uk/
https://lbh-frontend.herokuapp.com/components/lbh-skip-link

## Link to Jira card
https://hackney.atlassian.net/browse/MAAP-96

## Things to check
- Environment variables have been updated


app/views/layouts/application.html.erb
```diff
@@ -25,7 +25,7 @@
   <body data-no-turbolink="true">
     <%= render 'layouts/header' %>
     <%= render 'layouts/flash' %>
-    <main>
+    <main class="govuk-main-wrapper" id="main-content" role="main">
```
C-gyorfi, line 28:
```suggestion
    <main class="govuk-main-wrapper" role="main">
      <div id="phase banner" role="main">
        <%= render 'layouts/phase_banner' %>
        <div id="main-content">
          <%= yield %>
        </div>
      </div>
    </main>
```

C-gyorfi: The `html.erb` pages got rendered in the yield block, so if you'd add a div with the `main-content` id around this block the link works 👍  Something like the suggestion above

app/assets/stylesheets/components/skip_link.scss
```diff
@@ -0,0 +1,53 @@
+.skip_link {
```
C-gyorfi, line 1: As we discussed on the call, I'd cut down all the unnecessary tags here

app/views/layouts/application.html.erb
```diff
@@ -23,12 +23,17 @@
   </head>
 
   <body data-no-turbolink="true">
+    <div id="skip_link">
+      <a href="#main-content" class="skip_link">Skip to main content</a>
```
C-gyorfi, line 27:
```suggestion
      <a href="#main-content" data-turbolinks="false" class="skip_link">Skip to main content</a>
```

So I just found a way to fix the bug with the skip link, I've noticed anytime you'd click on the link the page is reloaded. We are using [Turbolinks](https://github.com/turbolinks/turbolinks) which is a tool intended to make page navigation more efficient, but it seems misconfigured, anytime we click on the link it refreshes the entire page. If you disable `turbolinks` on this link it works just fine and it wont reload the entire page

LBHLMatthewsMurray: Was wondering about this but wasn't too sure what Turbolinks were so it's great you found this - thanks for looking into it!

app/assets/stylesheets/components/main.scss
```diff
@@ -1,9 +1,10 @@
 main {
   background-color: $white;
-  padding: 1.5em 0;
+  //padding: 1.5em 0;
```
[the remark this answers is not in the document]
LBHLMatthewsMurray: Since changing around the order of the application.html.erb file there was extra padding being applied to the phase banner (I wasn't goo sure if I should remove this styling or not)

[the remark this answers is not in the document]
C-gyorfi: You can remove this if no longer needed!

PR #329, opened 2020-07-27
Create court case

Ninamma:
## Context

As part of building out formal agreements, we also want to also allow users to add court cases for a tenant through MA, since formal agreements link to a court case.

This is possibly the first slice of adding court cases to the front end. 🍰 

It is hidden behind a `create_formal_agreements` feature flag but is also nested inside the `create_informal_agreements` flag (since right now the button to create a new court case is in the agreements section of the case profile)

This PR:
- adds `create_formal_agreements` feature flag
image
- adds button to agreements section to create a court case on case profile page
image
- adds a basic page for creating a court case
image
- adds controller, use case and gateway to allow frontend to talk to income api 🗣️ 

Things to be done:
- Make relevant changes after we get designs for the court case UI
- Add functionality to allow users to view court cases
- Allow users to create and link formal agreements with court cases

## Guidance to review

This is heavily "inspired" by how we create agreements - and since we don't have court cases super fleshed out, just any thoughts would be appreciated!
- Not sure if adding a feature flag like I've done is what's best 🤔 
## Link to Jira card

https://hackney.atlassian.net/browse/MAAP-364?atlOrigin=eyJpIjoiMGI3YWJlYWQxMTUzNDEwYmJlOGNlZDMyNjU1MTI1YjIiLCJwIjoiaiJ9

## Things to check

### DIFF
--- a/app/views/court_cases/new.html.erb
+++ b/app/views/court_cases/new.html.erb
@@ -30,7 +30,7 @@
 
       <div class="form-group">
         <label class="govuk-label" for="balance_at_outcome_date"><strong>Balance at outcome date</strong><br/></label>
-        <%= number_field_tag(:balance_at_outcome_date, @balance_at_outcome_date, { class: 'form-control', required: true, min: 1, max: @tenancy.current_balance, placeholder: '£' }) %>
+        <%= number_field_tag(:balance_at_outcome_date, @balance_at_outcome_date, { class: 'form-control', required: true, min: 1, placeholder: '£' }) %>
       </div>
 
       <%= submit_tag 'Create', class: 'button' %>
--- a/app/views/court_cases/new.html.erb
+++ b/app/views/court_cases/new.html.erb
@@ -1,0 +1,39 @@
+<% content_for :title do %><%= @tenancy_ref %> - Create court case<% end %>
+
+<div class="grid-row">
+  <div class="column-full">
+    <%= link_to('Return to case profile', tenancy_path(id: @tenancy_ref), class: 'link--back') %>
+  </div>
+</div>
+
+<div class="grid-row">
+  <div class="column-full">
+    <h1><%= 'Create court case' %></h1>
+    <label class="govuk-label"><strong>Court case for: </strong><%= @tenancy.primary_contact_name %><br/></label>
+    <hr>
+  </div>
+</div>
+
+<div class="grid-row">
+  <div class="column-full">
+    <%= form_tag('create', method: :post) do %>
+
+      <div class="form-group">
+        <label class="govuk-date-field" for="court_decision_date"><strong>Court decision date</strong><br/></label>
+        <%= date_field_tag :court_decision_date, @court_decision_date, class: 'form-control' %>
+      </div>
+
+      <div class="form-group">
+        <label class="govuk-label" for="court_outcome"><strong>Court outcome</strong><br/></label>
+        <%= text_area_tag(:court_outcome, @court_outcome, { class: 'form-control' }) %>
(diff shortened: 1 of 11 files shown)


spec/features/feature_flags/manage_feature_flags_spec.rb
```diff
@@ -31,24 +32,33 @@ def when_i_visit_the_feature_flags_dashboard
 
   def then_i_should_see_the_existing_feature_flag
     expect(page).to have_content 'Create informal agreements Disabled'
+    expect(page).to have_content 'Create formal agreements Disabled'
   end
 
   def when_i_activate_the_feature
     click_button 'Activate Create informal agreements'
+    click_button 'Activate Create formal agreements'
```
C-gyorfi, line 40: Its not really an issue, but its not necessary to add each feature flags in this spec file, the purpose of this test is testing whether the feature flag system works, rather than the flags

Ninamma: yes, I did wonder that!!!! thank you for letting me know - I did think it was a bit too repetetive 😅

app/views/court_cases/new.html.erb
```diff
@@ -0,0 +1,39 @@
...
+      </div>
+
+      <div class="form-group">
+        <label class="govuk-label" for="court_outcome"><strong>Court outcome</strong><br/></label>
+        <%= text_area_tag(:court_outcome, @court_outcome, { class: 'form-control' }) %>
+      </div>
+
+      <div class="form-group">
+        <label class="govuk-label" for="balance_at_outcome_date"><strong>Balance at outcome date</strong><br/></label>
+        <%= number_field_tag(:balance_at_outcome_date, @balance_at_outcome_date, { class: 'form-control', required: true, min: 1, max: @tenancy.current_balance, placeholder: '£' }) %>
```
C-gyorfi, line 33: `max: @tenancy.current_balance` 

Is this rule true? I think this the balance at outcome date may be greater than the balance on the day?

app/views/tenancies/case/_agreements.html.erb
```diff
@@ -34,6 +34,10 @@
         <div class="column-full">
           <% agreement_button_title = @agreement.nil? ? 'Create agreement' : 'Cancel and create new' %>
           <%= link_to agreement_button_title, new_agreement_path(@tenancy.ref), class:'button' %>
+          <% if FeatureFlag.active?('create_formal_agreements') %>
```
C-gyorfi, line 37: This is currently also hidden by the `create_informal_agreements` feature flag, I wonder if a court outcome should have its own partial?

app/views/court_cases/new.html.erb
Ninamma: I think i just left it in there for now since i wasn't sure, but thanks for highlighting - i don't know if it might be figured out at the formal agreements meeting tomorrow 🤔

app/views/tenancies/case/_agreements.html.erb
Ninamma: I think so - I was planning to maybe split that out in the next 'view court cases' task, since right now the partial would just have a button

app/views/court_cases/new.html.erb
Ninamma: oh wait, I think i've been misunderstanding what that field is for 😅 

yeah you're right it totally could be greater than the balance on the day the court case is added

C-gyorfi: Sorry I wasn't very clear on this :D

PR #331, opened 2020-08-03
Update agreement view with breach data

C-gyorfi:
## Context
As a caseworker, I want to see the agreement state details so that I know when was the status checked by the system, what is the current state and why(e.g expected balance is lower than the checked balance).

- Update the `agreement status` partial to show `state`, `checked balance`, `expected balance` and `last checked` date

### Update view:
image

## Link to Jira card

https://hackney.atlassian.net/secure/RapidBoard.jspa?rapidView=30&projectKey=MAAP&modal=detail&selectedIssue=MAAP-175

## Things to check

### DIFF
--- a/spec/features/page_navigation_spec.rb
+++ b/spec/features/page_navigation_spec.rb
@@ -58,10 +58,6 @@ describe 'Page navigation' do
         }]
       }
     }.to_json
-
-    # stub_request(:get, 'https://example.com/tenancy/api/v1/tenancies/TEST%2F01/contacts')
-    #   .with(headers: { 'X-Api-Key' => ENV['TENANCY_API_KEY'] })
-    #   .to_return(status: 200, body: response_json)
   end
 
   def stub_income_api_my_cases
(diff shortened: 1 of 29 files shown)


spec/features/page_navigation_spec.rb
```diff
@@ -75,9 +59,9 @@ def stub_tenancy_api_contacts
       }
     }.to_json
 
-    stub_request(:get, 'https://example.com/tenancy/api/v1/tenancies/TEST%2F01/contacts')
-      .with(headers: { 'X-Api-Key' => ENV['TENANCY_API_KEY'] })
-      .to_return(status: 200, body: response_json)
+    # stub_request(:get, 'https://example.com/tenancy/api/v1/tenancies/TEST%2F01/contacts')
```
Ninamma, line 62: I guess these can just go?

C-gyorfi: Yep, good catch! 🤭

PR #332, opened 2020-08-03
Allow officers to add court date

Ninamma:
## Context

 </img>

- Slight changes to parameter names in Court Cases gateway, controller and use case to match the changes in the backend
  - `date_of_court_decision` -> `court_date`, `balance_at_outcome_date` -> `balance_on_court_outcome_date`
  - It's a little messy since the gateway, controller and use case aren't actually being used right now so they will need to be updated again once the endpoints in the backend have been changed 😓 

- Created Court Case partial for case profile - no longer nested inside Agreements partial

**Before**

 </img>

**After**
 </img>

- To match design and allow officers to add court dates before the court case takes place, replaced the `create court case` page with just a `add court date` page as the first step for adding a court case
 </img>

## Guidance to review

I don't think I've been making changes for court cases very neatly/gracefully - so I'm sorry! And happy to answer any questions if this is confusing.

## Link to Jira card

https://hackney.atlassian.net/browse/MAAP-349?atlOrigin=eyJpIjoiMGI3ZmM1MGU4NDFhNGZmNWIwOTU0Mjg3ZTc5YjRhNzYiLCJwIjoiaiJ9
## Things to check

### DIFF
--- a/app/views/tenancies/show.html.erb
+++ b/app/views/tenancies/show.html.erb
@@ -22,6 +22,7 @@
   <div class="column-two-thirds">
     <%= render :partial => 'tenancies/case/balance' %>
     <%= render :partial => 'tenancies/case/tenancy_attributes' %>
+    <%= render :partial => 'tenancies/case/court_cases' %>
     <%= render :partial => 'tenancies/case/agreements' %>
     <%= render :partial => 'tenancies/case/actions' %>
   </div>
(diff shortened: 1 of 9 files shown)


app/views/tenancies/show.html.erb
```diff
@@ -22,6 +22,7 @@
   <div class="column-two-thirds">
     <%= render :partial => 'tenancies/case/balance' %>
     <%= render :partial => 'tenancies/case/tenancy_attributes' %>
+    <%= render :partial => 'tenancies/case/court_cases' %>
```
C-gyorfi, line 25: Maybe you could have added the feature flag here, but it doesn't really matter 👍

PR #336, opened 2020-08-12
Action diary / timeline fix

elena-vi:
## Context

- When action diary code name is unknown, nothing is displayed
- Some action diary entries don't have messages, this causes the following error https://sentry.io/organizations/london-borough-of-hackney/issues/1836874406/?project=1451764&referrer=slack

- If action diary code name is unknown, display the code
- If action diary message is nil, set message to be an empty array

## Things to check

### DIFF
--- a/lib/hackney/income/action_diary_entry_codes.rb
+++ b/lib/hackney/income/action_diary_entry_codes.rb
@@ -205,10 +205,9 @@ module Hackney
       end
 
       def self.human_readable_action_code(code)
-        human_readable_action_code = all_code_options.find { |e| e.fetch(:code) == code }&.fetch(:name)
-        return code if human_readable_action_code.nil?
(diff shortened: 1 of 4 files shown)


lib/hackney/income/action_diary_entry_codes.rb
```diff
@@ -205,7 +205,10 @@ def self.valid_code?(test_code, user_accessible: nil)
       end
 
       def self.human_readable_action_code(code)
-        all_code_options.find { |e| e.fetch(:code) == code }&.fetch(:name)
+        human_readable_action_code = all_code_options.find { |e| e.fetch(:code) == code }&.fetch(:name)
```
antonyoneill, line 208: What does the `&` do in this syntax?

elena-vi: the `&` means that it will only run the method on right if the thing on the left is not nil

PR #354, opened 2020-08-24
Add missing court outcomes

C-gyorfi:
## Context
We are syncing these so we need to include them in the helper otherwise
we cannot display these correctly.
- Add missing court outcomes

### DIFF
--- a/app/helpers/court_outcomes_helper.rb
+++ b/app/helpers/court_outcomes_helper.rb
@@ -15,14 +15,17 @@ module CourtOutcomesHelper
 
   def court_outcomes
     {
+      'ADT' => 'Adjourned on Terms',
       'AGP' => 'Adjourned generally with permission to restore',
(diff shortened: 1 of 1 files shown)


app/helpers/court_outcomes_helper.rb
```diff
@@ -15,14 +15,17 @@ def court_outcomes_map
 
   def court_outcomes
     {
+      'AOT' => 'Adjourned on Terms',
```
Ninamma, line 18:
```suggestion
      'ADT' => 'Adjourned on Terms',
```
I think?

Ninamma: 'AOT' is ADJOURNED_ON_TERMS_SECONDARY? no idea what that is though 🤔

Ninamma (approved):
looks good! just the one comment - but not 100% sure about it

PR #355, opened 2020-08-25
Quick fix not to show old court details 

C-gyorfi:
## Context
There is a bug when we are checking if a court case has expired: it blows up when the court date is nil.
We don't want to render the old court cases anymore :feelsgood:  

- Fix court case expired bug
- Quick fix where we render old court cases we should render the new court case data once the feature flag is enabled

## Guidance to review
- We are going fast and introducing too many bugs as we don't have time to test edge cases 😨 

## Things to check
- [X] Environment variables have been updated


Ninamma (approved):
👍

PR #356, opened 2020-08-25
Fix adjourned on terms

Ninamma:
## Context

During the demo, we realised there was a bug when adding a 'Adjourned on terms' court outcome to a court case. 🐛

This was because we had recently added some more outcomes and codes to MA, but they were missing in the frontend's Court Case model

- Add new outcomes to Court Case model
- Remove old/outdated hash of court outcomes and codes

## Link to Jira card

https://hackney.atlassian.net/browse/MAAP-456?atlOrigin=eyJpIjoiZDE1ZmE5OWRiMDBkNDBiNDg2ZTMwOWRmZTI1MWUyNjAiLCJwIjoiaiJ9

### DIFF
--- a/app/helpers/court_outcomes_helper.rb
+++ b/app/helpers/court_outcomes_helper.rb
@@ -1,4 +1,18 @@
 module CourtOutcomesHelper
+  def court_outcomes_map
+    {
+      'AAH' => 'Adjourned to another hearing date',
+      'AGE' => 'Adjourned generally',
+      'AOT' => 'Adjourned on Terms',
+      'OPD' => 'Outright Possession (with Date)',
+      'OPF' => 'Outright Possession (Forthwith)',
+      'PPO' => 'Postponed Possession Order',
+      'STR' => 'Struck out',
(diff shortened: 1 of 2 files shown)


app/helpers/court_outcomes_helper.rb
```diff
@@ -1,18 +1,4 @@
 module CourtOutcomesHelper
-  def court_outcomes_map
```
C-gyorfi, line 2: Are we using this anywhere else? Where we render the old court outcomes?

C-gyorfi (approved):

app/helpers/court_outcomes_helper.rb
Ninamma: Good shout - I had previously thought it would be fine, but I guess it would be good to keep in case we need to rollback like for here - I'll update
image

C-gyorfi: Yep lets make this change when we remove the feature flag, hopefully before the end of the extension 🤞

PR #363, opened 2020-09-03
Display agreement end date on agreement summary

Miles-Alford:
## Context

Based on feedback from the users, it was requested that the end date of an agreement be added onto the agreement summary section on a tenancy page so they do not have to change page to the view details page to view this information.

- Add the agreement end date to the agreement status section 

### Before
image

### After
image

## Guidance to review

Not really sure where in the box they would like to view this information so I put it to the left of the agreements status, recommendations would be nice if there is a better positioning.

## Link to Jira card

https://hackney.atlassian.net/browse/MAAP-471

## Things to check


C-gyorfi (approved):
👍

PR #367, opened 2020-09-08
display starting balance while creating a formal agreement

elena-vi:
## Context

formal agreements should have the same starting balance as the balance at court outcome and this should be visible to the user
Screenshot 2020-09-08 at 11 32 36

- display the starting balance which is the same as the balance at court outcome
- send the starting balance to the api 

## Link to Jira card

https://hackney.atlassian.net/secure/RapidBoard.jspa?rapidView=30&modal=detail&selectedIssue=MAAP-468

## Things to check

### DIFF
--- a/app/views/agreements/form/_agreement_formal.html.erb
+++ b/app/views/agreements/form/_agreement_formal.html.erb
@@ -3,7 +3,7 @@
     <legend>
       <h3>Court case related to this agreement</h3> <br>
     </legend>
-    
+
     <%= hidden_field_tag :court_case_id, @court_cases.last.id %>
     <label for=<%= "court_case_id_#{@court_cases.last.id}" %>>
       <strong>Court date:</strong> <%= format_date(@court_cases.last.court_date) %>
@@ -12,6 +12,10 @@
   </fieldset>
 </div>
 
+<div class="form-group">
+  <label class="govuk-label" style="font-weight:bold" for="starting_balance" id="starting_balance_label">Starting Balance</label><br/>
+  <%= number_field_tag :starting_balance, @court_cases.last.balance_on_court_outcome_date, { class: 'form-control', required: true, readonly: true} %>
+</div>
 
 <%= render :partial => "agreements/form/agreement_informal" %>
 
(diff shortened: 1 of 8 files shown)


Ninamma (approved):
👍

app/views/agreements/form/_agreement_formal.html.erb
```diff
@@ -12,6 +12,10 @@
   </fieldset>
 </div>
 
+<div class="form-group">
+  <label class="govuk-label" style="font-weight:bold" for="starting_balance" id="starting_balance_label">Starting Balance</label><br/>
+  <%= number_field_tag :starting_balance, @court_cases.last.balance_on_court_outcome_date, { class: 'form-control', required: true, readonly: true} %>
```
C-gyorfi, line 17: Why is this an editable field? I thought the starting balance in a court agreement is the balance at court outcome date?

lib/hackney/income/create_agreement.rb
```diff
@@ -5,7 +5,7 @@ def initialize(agreement_gateway:)
         @agreement_gateway = agreement_gateway
       end
 
-      def execute(tenancy_ref:, agreement_type:, frequency:, amount:, start_date:, created_by:, notes:, court_case_id:)
+      def execute(tenancy_ref:, agreement_type:, frequency:, amount:, start_date:, created_by:, notes:, court_case_id:, starting_balance: nil)
```
C-gyorfi, line 8: Maybe we don't need this, because the create formal agreement use-case in API looks up the balance on case details. Although this is currently wrong, but we could change the API to grab the balance from the court case. Any thoughts?

elena-vi: I just came to realise this while doing the backend fix. 
Going to finish the backend first then amend this one

app/views/agreements/form/_agreement_formal.html.erb
elena-vi: it's not editable, its readonly

C-gyorfi (approved):
👍

PR #369, opened 2020-09-08
Allow viewing an agreement with variable payments

C-gyorfi:
## Context
As a caseworker, I want to be able to see when an agreement is a variable payment agreement, so I can see when the initial larger payment is expected to be paid. 

- Update agreement gateway to pull variable payment fields
- Allow showing variable payment details on agreement details and history pages 

image

## Link to Jira card
https://hackney.atlassian.net/secure/RapidBoard.jspa?rapidView=30&projectKey=MAAP&modal=detail&selectedIssue=MAAP-464

## Things to check


Ninamma (approved):
looks good, the only thing i can think of is the language change from standup, so `one-off` -> `lump sum` 🤔

PR #371, opened 2020-09-09
mark formal agreements as court ordered agreements

elena-vi:
## Context

When a user is looking at an agreement, it is not clear when that agreement is court ordered
Screenshot 2020-09-09 at 11 25 41
Screenshot 2020-09-09 at 11 25 35

- When agreement is formal, the agreement status will be shown as `Court ordered` 
- Display the agreement type in history

## Things to check


C-gyorfi (approved):
🚀

PR #372, opened 2020-09-09
New new agreement button

elena-vi:
Screenshot 2020-09-09 at 14 38 22

## Context

make sure that the user knows that they will be created a new court agreement if they cancel one

## Things to check


Ninamma (approved):
👍

PR #374, opened 2020-09-10
formal agreement bug fix

elena-vi:
## Context

When there is an existing formal agreement and you try creating informal agreement, it fails because this shouldn’t be allowed

when there is a formal agreement show cancel agreement when new court case is added

## Things to check


C-gyorfi (approved):
Smart, I love it! :shipit:

PR #375, opened 2020-09-10
implement infinite scrolling on case page

elena-vi:
## Context

the case pages take very long to load, I suspect because it takes too long to render the action diary timeline, implementing infinite scrolling might make the case pages load faster and provide the same user experience 

implement infinite scroll on the timeline. 

When a user scrolls to the bottom of the page, they will see the following message followed by more entries

Screenshot 2020-09-10 at 13 12 08

## Things to check


C-gyorfi (approved):
Works like a charm 🚀

PR #378, opened 2020-09-15
Allow adding court times

Ninamma:
## Context

We want to allow caseworkers to be able to add the time of an upcoming court hearing - this is mainly for it to be used for the Court Date letter.

- Added field in Court Date form for `court_time`
- Display `court_time` on the Court Case details page

To do:
- ~~Figure out if Time Zones matter~~ **I think they don't**
  - ~~If I'm currently in BST and add a court date that would be in GMT, will that affect the court time when we are in GMT?~~
  - ~~Is there a way to just ignore Time Zones?~~
  - **I think we don't need to do anything about them, since we just use `strftime` to turn DateTimes into strings and don't do anything with the Time Zones at all.**

- Check if the Court time needs to be displayed anywhere else
- Update CourtDate in API docs

## Guidance to review

I realised while going around in circles to find a way to input time that my Rails frontend knowledge isn't so good - so any pointers there would be good!

Let me know if what I've said about Time Zones makes sense!
## Link to Jira card

https://hackney.atlassian.net/browse/MAAP-487?atlOrigin=eyJpIjoiNDA2ZTMzYWE4ZDM1NGZmN2E3YTk3NTMwZTQwZmQzNDQiLCJwIjoiaiJ9

## Things to check

### DIFF
--- a/app/controllers/court_cases_controller.rb
+++ b/app/controllers/court_cases_controller.rb
@@ -8,9 +8,8 @@ class CourtCasesController < ApplicationController
   def create
     create_court_case_params = {
       tenancy_ref: tenancy_ref,
-      court_date: params.fetch(:court_date)
+      court_date: court_date
     }
-
     use_cases.create_court_case.execute(create_court_case_params: create_court_case_params)
 
     redirect_to show_success_court_case_path(message: 'Successfully created a new court case')
@@ -22,6 +21,7 @@ class CourtCasesController < ApplicationController
   def edit_court_date
     @tenancy = use_cases.view_tenancy.execute(tenancy_ref: tenancy_ref)
     @court_date = court_case.court_date&.to_date&.strftime('%F')
+    @court_time = court_case.court_date&.to_time&.strftime('%R')
   end
 
   def update_court_date
@@ -131,5 +131,11 @@ class CourtCasesController < ApplicationController
   def to_boolean(param)
     return true if param == 'Yes'
     return false if param == 'No'
+  end
+
+  def court_date
+    court_date = params.fetch(:court_date)
+    court_time = params.fetch(:court_time)
+    "#{court_date} #{court_time}"
   end
 end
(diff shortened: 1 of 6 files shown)


C-gyorfi (commented):
when I edited an existing court time 😱 
image

app/controllers/court_cases_controller.rb
```diff
@@ -22,6 +21,7 @@ def create
   def edit_court_date
     @tenancy = use_cases.view_tenancy.execute(tenancy_ref: tenancy_ref)
     @court_date = court_case.court_date&.to_date&.strftime('%F')
+    @court_time = court_case.court_date&.to_time&.strftime('%R')
   end
 
   def update_court_date
```
C-gyorfi, line 27: I think we need to update this action too, currently the update always generates a `00:00` time

Ninamma: Fixing this now, but am so confused as to why the test pass - line 176 in `creating_court_case_spec` should fail in this case 🙃 

image

spec/features/income_collection/court_cases/creating_court_case_spec.rb
```diff
@@ -35,8 +35,8 @@
 
     when_i_click_on_edit_court_date
     then_i_should_see_edit_court_date_page
-    and_i_should_see_the_current_curt_date
-    when_i_fill_in_the_new_court_date
+    and_i_should_see_the_current_court_date_info
+    when_i_fill_in_the_new_court_date_and_time
     and_i_click_on_save
 
     then_i_should_see_the_tenancy_page
```
C-gyorfi, line 42: Yep, could we could have changed the expectation when updating the court date in `and_i_should_see_the_updated_court_date`

C-gyorfi (approved):
👍

spec/features/income_collection/court_cases/creating_court_case_spec.rb
Ninamma: Oh no, so that had been updated. 

What I did was I had a stubbed API response with the updated time - so that expectation still passed without a working controller.
But in the production code I hadn't actually made the controller send the correct info to the API 😬 danger of stubbing without thinking 😓

PR #379, opened 2020-09-16
Removing random 'thing'

Ninamma:
get rid of thing:
image


C-gyorfi (approved):
```git blame app/views/court_cases/show.html.erb``` 😃

elena-vi (commented):
that was me right? 
sounds like some**thing** I would do

C-gyorfi:
Yes @elena-vi if it was me, it would be something vulgar

PR #382, opened 2020-09-21
Allow creating formal one off payment agreements

C-gyorfi:
## Context
The one-off payment form does not allow creating formal agreements because the court case id is missing from the form.

- Add `court_case_id` as a hidden field tag when it's a formal agreement

## Things to check


Ninamma (approved):
👍

PR #385, opened 2020-09-22
Ask users to fill in the reason of cancellation when cancelling an agreement

C-gyorfi:
## Context
As a credit controller, whenever I cancel or change an agreement I want to be able to add a note so that I can put in a reason for why it was cancelled or changed.

- Update `create agreement` use-case and `agreements gateway` to send `cancellation reason` and `cancelled by` in the request body
- Ask users to fill in the reason of cancellation when cancelling an agreement
image

## Link to Jira card
https://hackney.atlassian.net/secure/RapidBoard.jspa?rapidView=30&projectKey=MAAP&modal=detail&selectedIssue=MAAP-280
## Things to check

### DIFF
--- a/app/controllers/agreements_controller.rb
+++ b/app/controllers/agreements_controller.rb
@@ -60,16 +60,21 @@ class AgreementsController < ApplicationController
   def confirm_cancellation
     tenancy_ref
     agreement_id
+    cancellation_reason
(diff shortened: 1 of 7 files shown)


Ninamma (approved):
👍

PR #387, opened 2020-09-23
Court outcome action diary

Ninamma:
## Context

When an officer adds a Court Outcome to a Court case, we want an entry written to the action diary and the case to be resynced.

This is the first slice for this 🍰 

- Get frontend to pass `username` and `tenancy ref` to `update court case` use case in the back end
## Link to Jira card

https://hackney.atlassian.net/browse/MAAP-522?atlOrigin=eyJpIjoiODQ3MmZjMTI4NDQ0NDRmZDhiMTA0OTE3MjdhNzRiMjAiLCJwIjoiaiJ9
## Things to check

### DIFF
--- a/app/controllers/court_cases_controller.rb
+++ b/app/controllers/court_cases_controller.rb
@@ -51,7 +51,8 @@ class CourtCasesController < ApplicationController
       id: court_case_id,
       court_outcome: court_outcome,
       balance_on_court_outcome_date: params.fetch(:balance_on_court_outcome_date),
-      strike_out_date: params.fetch(:strike_out_date)
+      strike_out_date: params.fetch(:strike_out_date),
+      username: username
     }
 
     if Hackney::Income::Domain::CourtCase.new(court_outcome: court_outcome).can_have_terms?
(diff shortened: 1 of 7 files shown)


app/controllers/court_cases_controller.rb
```diff
@@ -82,7 +84,9 @@ def update_terms
       disrepair_counter_claim: to_boolean(params.fetch(:disrepair_counter_claim)),
       court_outcome: params.fetch(:court_outcome),
       balance_on_court_outcome_date: params.fetch(:balance_on_court_outcome_date),
-      strike_out_date: params.fetch(:strike_out_date)
+      strike_out_date: params.fetch(:strike_out_date),
+      username: username,
+      tenancy_ref: tenancy_ref
```
C-gyorfi, line 89: I may be wrong, but I think we don't need the `tenancy_ref` param

Ninamma: When you say we don't need it, do you mean we don't need it because there's an easier way to get it than passing it like this? 

Or we don't need it at all? 🤔 

Because from what I can see, the usecase to write into the Action Diary and resync does require it:
```
@add_action_diary_and_sync_case_usecase.execute(
            username: username,
            tenancy_ref: tenancy_ref,
            action_code: 'IC6',
            comment: "Court outcome added: #{human_readable_court_outcome}"
          )
```

C-gyorfi: Sorry I wasn't clear, so a court case has tenancy reference:
```RUBY
  create_table "court_cases", options: "ENGINE=InnoDB DEFAULT CHARSET=utf8", force: :cascade do |t|
...
    t.string "tenancy_ref", null: false
...
  end
```

Ninamma: ah yes, yeah that makes a lot of sense! thank you! will sort that out! :D

C-gyorfi (approved):
🚀

PR #388, opened 2020-09-24
Add covid pause reason

Ninamma:
## Context

Officers might need to pause cases, where there is a Court Order breach because of Covid.

This is the frontend change to allow it.

I also realised that I hadn't added the action codes for `Court Date Letter` and `Court Outcome Added` to the list of valid Action Codes.

- Adds Pause dropdown option for `C19 Court Order Breached`

- Adds Action Codes for `C19 Court Order Breached`, `Court Date Letter` and `Court Outcome Added` to list of Action Diary Entry Code Options
## Guidance to review

I'm not 100% confident with Action Diary related things, so please shout if it looks like I'm missing anything 😅 
## Link to Jira card

https://hackney.atlassian.net/browse/MAAP-519?atlOrigin=eyJpIjoiZTgwMzkzZjUxMmVkNDE2ZGE0ZjI1YWQ4NTc4NWVmMTYiLCJwIjoiaiJ9
## Things to check

### DIFF
--- a/app/helpers/pause_reason_filter_helper.rb
+++ b/app/helpers/pause_reason_filter_helper.rb
@@ -8,6 +8,7 @@ module PauseReasonFilterHelper
         'Promise of payment' => 'Promise of payment',
         'Deceased' => 'Deceased',
         'Missing Data' => 'Missing Data',
+        'C19 Court Order Breached' => 'C19 Court Order Breached',
         'Other' => 'Other'
       }, option
     )
--- a/app/helpers/tenancy_helper.rb
+++ b/app/helpers/tenancy_helper.rb
@@ -20,6 +20,7 @@ module TenancyHelper
       'Promise of payment' => 'POP',
       'Deceased' => 'DEC',
       'Missing Data' => 'RMD',
+      'C19 Court Order Breached' => 'CVB',
       'Other' => 'GEN'
     }
   end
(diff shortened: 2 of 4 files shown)


app/helpers/pause_reason_filter_helper.rb
```diff
@@ -8,6 +8,7 @@ def pause_reason_filter(option)
         'Promise of payment' => 'Promise of payment',
         'Deceased' => 'Deceased',
         'Missing Data' => 'Missing Data',
+        'C19 Court Order Breached' => 'C19 Court Order Breached',
```
antonyoneill, line 11: Some criticism of the existing pattern here. This file maps keys to values, where keys and values are equal. In reality, we should probably be using the reason codes as the keys here.

1/2

app/helpers/tenancy_helper.rb
```diff
@@ -20,6 +20,7 @@ def pause_reasons
       'Promise of payment' => 'POP',
       'Deceased' => 'DEC',
       'Missing Data' => 'RMD',
+      'C19 Court Order Breached' => 'CVB',
```
antonyoneill, line 23: I'd make this the source of truth (or a separate file) of `code` -> `Human readable` key values, and use this directly in the above.

2/2

antonyoneill (approved):
This is OK as is, my comments are meant as a prompt for challenging existing patterns - this isn't the time to address them

app/helpers/pause_reason_filter_helper.rb
Ninamma: Yeah, I did find that odd - but went with ye old `I'm sure it's done like that for a reason so I won't question it` thinking - but will avoid that in the future 😅

PR #391, opened 2020-09-24
Remove duplicate pause reason definitions

C-gyorfi:
## Context
Following the conversation on this [PR #388](https://github.com/LBHackney-IT/LBH-IncomeCollection/pull/388/files/922a664cd8cdc698f3101b25e48f0c3d23ee17ad#)
Remove duplicate pause reason definitions


app/helpers/tenancy_helper.rb
```diff
@@ -25,6 +25,12 @@ def pause_reasons
     }
   end
 
+  def pause_reason_filter(option)
```
antonyoneill, line 28: How is this being referenced?

app/helpers/pause_reason_filter_helper.rb
```diff
@@ -1,16 +0,0 @@
-module PauseReasonFilterHelper
```
antonyoneill, line 1: Do we need to replace the import of this helper with the tenancy helper? or does rails auto import this?

app/helpers/tenancy_helper.rb
C-gyorfi: its referenced in the `_pause_reason_filter.html.erb` 
```RUBY
          <%= select_tag(:pause_reason, pause_reason_filter(@filter_params.pause_reason), { :class => 'form-control', prompt: 'Select a pause reason'  }) %>
```
 tenancy_helper.rb  is loaded trough rails_helper.rb as any other helpers, apparently its accecible in any erb file

antonyoneill: Great! Thanks - that's why we don't need to remove any references to the PauseReasonFilterHelper

antonyoneill (approved):

PR #397, opened 2020-10-09
Create send and view letter buttons and remove leasehold buttons from rents worktray

Miles-Alford:
## Context

Small PR which adds two new buttons for the rents worktray to send bulk letters and to view them.
Also removes the leasehold buttons from the same worktray that appear when you are within both user groups

- Add send & view letters button partial to allow rents to go straight to the bulk letter sending tool
- Remove displaying of leasehold buttons on the worktray

## Guidance to review

I have kept the leasehold button just in case we wish to use it within the V2 worktray however if not will delete it. Also a bit skeptical on the value of the worktray request spec tests, added a test nonetheless to keep to standard however I am unsure of the value of these tests
## Things to check

### DIFF
--- a/app/views/tenancies/income_collection/_buttons.html.erb
+++ b/app/views/tenancies/income_collection/_buttons.html.erb
@@ -1,0 +1,8 @@
+<div class="grid-row">
+  <div class="column-full">
+    <div class='other_actions tenancy_list column-align-center'>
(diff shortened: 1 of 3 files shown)


Duslerke (approved):

Duslerke:
Can you provide a screenshot of an issue?

